PR #15, opened 2022-10-06
Add balancer

vjerci:
🎁 What
---
- Add balancer
- Add git


pkg/commands/app.go
```diff
@@ -174,8 +193,8 @@ func NewScanCommand(globalFlags *flag.GlobalFlagGroup) *cobra.Command {
 			}
 			return artifact.Run(cmd.Context(), options, artifact.TargetFilesystem)
 		},
-		SilenceErrors: true,
-		SilenceUsage:  true,
+		SilenceErrors: false,
+		SilenceUsage:  false,
```
cfabianski, line 197: Should this be committed?

vjerci: I'll fix it once everything works. At this point I'm not even sure what it does.

PR #20, opened 2022-10-07
Json report

vjerci:
🎁 What
---
- Add support for json output
- Removed processing-worker from config list
- Made scan return error instead of starting when target isn't present

### DIFF
--- a/pkg/commands/app.go
+++ b/pkg/commands/app.go
@@ -101,10 +101,11 @@ func NewProcessingServerCommand() *cobra.Command {
 				return fmt.Errorf("options binding error: %w", err)
 			}
 
-			log.Debug().Msgf("got port `%s`", options.Port)
+			log.Debug().Msgf("running scan worker on port `%s`", options.Port)
 
 			return worker.Start(options.Port)
 		},
+		Hidden:        true,
 		SilenceErrors: true,
 		SilenceUsage:  true,
 	}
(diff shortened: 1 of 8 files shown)


pkg/commands/app.go
```diff
@@ -212,7 +214,9 @@ func NewScanCommand(globalFlags *flag.GlobalFlagGroup) *cobra.Command {
 				return xerrors.Errorf("flag error: %w", err)
 			}
 
-			log.Debug().Msgf("process online timeout is  %d", options.TimeoutWorkerOnline)
+			if options.Target == "" {
+				return fmt.Errorf("path is required")
```
cfabianski, line 218: Does it return an error or does it show the helper for the scan command?

pkg/commands/app.go
```diff
@@ -101,10 +101,11 @@ func NewProcessingServerCommand() *cobra.Command {
 				return fmt.Errorf("options binding error: %w", err)
 			}
 
-			log.Debug().Msgf("got port `%s`", options.Port)
+			log.Debug().Msgf("running scan worker on port `%s`", options.Port)
 
 			return worker.Start(options.Port)
 		},
+		Hidden:        true,
```
cfabianski, line 108: Nice

pkg/commands/app.go
vjerci: It shows both error and full  usage for sacn

```
Error: path is required
Usage:
  .....
  ```

cfabianski (approved):

PR #22, opened 2022-10-10
Classify dependencies

elsapet:
🎁 What
---
* Update recipes.json with packages from AMA database
* Add package manager to `dependencies.Dependency`
  * Update related dependency detectors and test snapshots
* Match against recipe packages when classifying dependency detections

✨ Why
---
Support package classification for dependency detections

### DIFF
--- a/pkg/detectors/dependencies/depsbase/depsbase.go
+++ b/pkg/detectors/dependencies/depsbase/depsbase.go
@@ -13,6 +13,6 @@ type Dependency struct {
 type DiscoveredDependency struct {
 	Provider       string       `json:"provider"`
 	Language       string       `json:"language"`
-	PackageManager string       `json:"package_manager`
+	PackageManager string       `json:"package_manager"`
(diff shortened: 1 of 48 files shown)


pkg/detectors/dependencies/depsbase/depsbase.go
```diff
@@ -11,7 +11,8 @@ type Dependency struct {
 
 // DiscoveredDependency holds a list of dependencies defined in package file
 type DiscoveredDependency struct {
-	Provider     string       `json:"provider"`
-	Language     string       `json:"language"`
-	Dependencies []Dependency `json:"dependencies"`
+	Provider       string       `json:"provider"`
+	Language       string       `json:"language"`
+	PackageManager string       `json:"package_manager`
```
cfabianski:
```suggestion
	PackageManager string       `json:"package_manager"`
```

cfabianski (commented):
So far so good!

cfabianski (approved):

PR #23, opened 2022-10-11
Add support for -skip-path flag

vjerci:
🎁 What
---
- Removed filler line for empty reports
- Refractor scanning for list of files
- Add support for skip-path flag
    - supports * syntax in skip-path

📚 example:
---

If you had directory structure
```
    admin.go
    users.go
    superusers/
        headmaster.go
```

and want to ignore `users.go` and `superusers dir` you could do 
```
curio scan ${target} --skip-path users.go,superusers
```

Basically you specify files you want to ignore same way as you do in gitignore but instead of separating them with `\n` you separate with `,`


cfabianski (commented):
Can you update the readme with the new flags/options please?

vjerci:
> Can you update the readme with the new flags/options please?

yeah its this ticket:
https://bearer.atlassian.net/jira/software/projects/AMA/boards/57?selectedIssue=AMA-2976

do you want me to ditch gorelease for now and work on that?

gotbadger:
@vjerci we need to have documenation alongside the code so please add it as part of this PR

cfabianski:
Let's check with @markmichon @gotbadger and @b1pb1p what they think about that flag. 
We could have `skip-directories` and `skip-files` but that feels a duplication. I'd rather have one single flag for it but I'm not a big fan of just `skip`. Maybe `skip-files` is enough (we could express paths in `skip-files` ala [.gitignore](https://git-scm.com/docs/gitignore) and it would accept patterns).

cfabianski (approved):

PR #29, opened 2022-10-12
AMA-2969 classify recipe urls

elsapet:
🎁 What
---
Classify interfaces that match recipe URLs

✨ Why
---
Support recipe classification for interface detections


pkg/classification/db/recipes/stripe.json
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+{"metadata":{"version":"1.0"},"name":"Stripe","type":"external_service","urls":["https://api.stripe.com", "https://js.stripe.com"],"packages":[{"name":"stripe","group":null,"package_manager":"rubygems"},{"name":"github.com/stripe/stripe-go","group":null,"package_manager":"go"},{"name":"stripe","group":null,"package_manager":"packagist"},{"name":"stripe","group":null,"package_manager":"pypi"},{"name":"stripe","group":null,"package_manager":"npm"},{"name":"react-stripe-elements","group":null,"package_manager":"npm"},{"name":"@stripe/stripe-js","group":null,"package_manager":"npm"},{"name":"@stripe/react-stripe-js","group":null,"package_manager":"npm"}]}
```
elsapet: This is for testing only; I add URLs for all of the recipes here: https://github.com/Bearer/curio/pull/42

cfabianski (dismissed):

pkg/classification/interfaces/interfaces.go
```diff
@@ -1,42 +1,142 @@
...
-		Detection: &data,
-		Classification: &Classification{
-			RecipeName: "stripe",
-		},
+func (classifier *Classifier) Classify(data report.Detection) (*ClassifiedInterface, error) {
+	var classifiedInterface *ClassifiedInterface
+	var classificationDecision ClassificationDecision
+
+	// detected url, with unknown parts replaced with * wildcards
+	value := data.Value.(interfaces.Interface).Value.ToString()
```
vjerci: feels like there should be some cast error checking here. because if casting fails it will panic
```
value, ok  := data.Value.(interfaces.Interface).Value.ToString()
if !ok {
   return nil, errors.New("passed detection is not interface")
}
```

pkg/classification/interfaces/interfaces.go
```diff
@@ -1,42 +1,142 @@
...
+	recipeMatch, err := classifier.FindMatchingRecipeUrl(value)
+	if err != nil {
+		return classifiedInterface, err
+	}
+
+	if recipeMatch != nil {
```
vjerci, line 108: feels like else can be swapped with early return to avoid tree branching
```
                 classifiedInterface := &ClassifiedInterface{
			Detection: &data,
			Classification: &Classification{
				URL:         recipeMatch.DetectionURLPart,
				RecipeMatch: true,
				RecipeName:  recipeMatch.RecipeName,
			},
		}

		if strings.Contains(recipeMatch.DetectionURLPart, "*") {
			classifiedInterface.Classification.Decision = ClassificationDecision{
				State:  Potential,
				Reason: "recipe_match_with_wildcard",
			}
			return classifiedInterface, nil
		} 
		
		classifiedInterface.Classification.Decision = ClassificationDecision{
			State:  Valid,
			Reason: "recipe_match",
		}
		return classifiedInterface, nil

```

pkg/classification/interfaces/interfaces.go
```diff
@@ -1,42 +1,142 @@
...
+	for _, recipe := range classifier.config.Recipes {
+		for _, recipeURL := range recipe.URLS {
+			match, err := url_matcher.UrlMatcher(
+				url_matcher.ComparableUrls{
+					DetectionURL: detectionURL,
+					RecipeURL:    recipeURL,
+				},
+			)
+			if err != nil {
+				return recipeURLMatch, err
```
vjerci: shouldn't this be `return nil, err`?

pkg/classification/interfaces/interfaces.go
vjerci: feels like this should be 
```
value, ok  := data.Value.(interfaces.Interface).Value.ToString()
if !ok {
    return nil, errors.New("detection is not interface")
}
```

pkg/util/url_matcher/url_matcher.go
```diff
@@ -0,0 +1,71 @@
...
+	parsedDomain, err := publicsuffix.ParseFromListWithOptions(
+		publicsuffix.DefaultList,
+		parsedURL.Host,
+		&publicsuffix.FindOptions{IgnorePrivate: true, DefaultRule: nil},
+	)
+	if err != nil {
+		return "", err
+	}
+
+	matcher, _ := regexp.Compile(prefixPattern + domainPattern(parsedDomain) + pathPattern(parsedURL) + ")" + suffixPattern)
```
vjerci: compiling regexp is expensive, think we should do this only once and not for each time we are trying to match something.

cfabianski (dismissed):

pkg/util/url_matcher/url_matcher.go
```diff
@@ -0,0 +1,71 @@
+package url_matcher
```
vjerci: would be better if package name is url and method is Match

pkg/classification/interfaces/interfaces.go
```diff
@@ -36,35 +37,63 @@ type Classification struct {
...
+	var preparedRecipes []Recipe
+	for _, recipe := range config.Recipes {
+		preparedRecipe := Recipe{
+			Name: recipe.Name,
+			Type: recipe.Type,
+		}
+		for _, recipeURL := range recipe.URLS {
+			regexpMatcher, err := url.PrepareRegexpMatcher(recipeURL)
+			if err != nil {
+				panic(err) // todo: how to handle error in New()?
```
elsapet: @vjerci I'm not sure how to handle an error in an initialiser - any suggestions?

go.sum
```diff
@@ -226,6 +228,10 @@ github.com/struCoder/pidusage v0.2.1 h1:dFiEgUDkubeIj0XA1NpQ6+8LQmKrLi7NiIQl86E6
 github.com/struCoder/pidusage v0.2.1/go.mod h1:bewtP2KUA1TBUyza5+/PCpSQ6sc/H6jJbIKAzqW86BA=
 github.com/subosito/gotenv v1.4.1 h1:jyEFiXpy21Wm81FBN71l9VoMMV8H8jG+qIK3GCpY6Qs=
 github.com/subosito/gotenv v1.4.1/go.mod h1:ayKnFf/c6rvx/2iiLrJUk1e6plDbT3edrFNGqEflhK0=
+github.com/weppos/publicsuffix-go v0.12.0/go.mod h1:z3LCPQ38eedDQSwmsSRW4Y7t2L8Ln16JPQ02lHAdn5k=
+github.com/weppos/publicsuffix-go v0.20.0 h1:59ypvSUbW3Dunc6zVm+v+MmXf2Q6cGiNDkxgRIzEnaA=
+github.com/weppos/publicsuffix-go v0.20.0/go.mod h1:5ZC/Uv3fIEUE0eP6o9+Yg4+5+W8V0/BieMi05feGXVA=
```
cfabianski, line 233: You probably need to run the command `go mod tidy` which will clean up your outdated dependencies
https://go.dev/ref/mod#go-mod-tidy

pkg/classification/interfaces/interfaces.go
vjerci: @elsapet 
same as any other function `return nil, err` it's a standard golang pattern,
for now just propagate it to consumer.

might be good to have a default error defined at package level, since we plan to eventually allow extending our db
```
var ErrInvalidRecepies = errors.New("invalid interfaces recipe")
func New(config Config) (*Classifier, error) {
   regexpMatcher, err := url.PrepareRegexpMatcher(recipeURL)
   if err != nil {
       return nil, ErrInvalidRecepies
   }
}
```

vjerci (approved):

PR #46, opened 2022-10-14
Classify invalid urls

elsapet:
## Description
Classify URLs as invalid, potential, or valid based on their structure 
This classification is overridden in the case of a recipe match

Still pending (for a future PR)
- Reachability check

## Checklist

### DIFF
--- a/pkg/util/url/url.go
+++ b/pkg/util/url/url.go
@@ -2,18 +2,192 @@ package url
 
 import (
 	"errors"
+	"net"
 	"net/url"
+	"path/filepath"
 	"regexp"
 	"strings"
 
+	"github.com/bearer/curio/pkg/report"
+	"github.com/bearer/curio/pkg/report/detectors"
 	"github.com/weppos/publicsuffix-go/publicsuffix"
 )
 
 const prefixPattern = "(?P<match>\\A(?:[^:]+://)?(?:[^/]+\\.)?"
 const suffixPattern = "(?:/|\\z)"
 
+// recipe url matching regexp
 var regexpReplaceMatcher = regexp.MustCompile(`<\w+>`)
 var regexpVariableMatcher = regexp.MustCompile(`\A[*\/.-:]+\z`)
+
+// url validation regexp
+var regexpDependencyFileMatcher = regexp.MustCompile(`Gemfile\.lock|package\.json|yarn\.lock|maven\-dependencies\.json|gemnasium\-maven\-plugin\.json|gradle\-dependencies\.json|Pipfile\.lock|package\-lock\.json|npm\-shrinkwrap\.json|packages\.lock\.json|project\.json|packages\.config|paket\.dependencies|ivy\-report\.xml|composer\.lock|composer\.json|pipdeptree\.json|go\.sum|requirements\.txt|pyproject\.toml|poetry\.lock|pom\.xml|build\.gradle`)
+var regexpInvalidFilenameMatcher = regexp.MustCompile(`(trad|/translations?/|locales?|dockerfile|i18n)`)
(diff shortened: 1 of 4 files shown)


cfabianski (approved):
Great work!

pkg/util/url/url.go
```diff
@@ -67,6 +241,160 @@ func PrepareURLValue(myURL string) (string, error) {
...
+	if isVendored(data.Source.Filename) {
+		ValidationResult.Reason = "included_in_vendor_folder"
+		return &ValidationResult, nil
+	}
+
+	if isPotentialDetector(data.DetectorType) {
+		ValidationResult.State = Potential
+		ValidationResult.Reason = "potential_detectors"
+		return &ValidationResult, nil
+	}
```
elsapet, line 260: We probably want to make these checks for all detection types (at least, it is this way in the Ruby app) - not sure the best place to put them though 🤔 
Any thoughts @vjerci @cfabianski

pkg/util/url/url.go
```diff
@@ -93,3 +421,61 @@ func wildcardPattern(myString string, separator string) string {
 	}
 	return strings.Join(parts, regexp.QuoteMeta(separator))
 }
+
+func domainIsExcluded(host string) bool {
+	_, ok := excludedDomains[host]
+	return ok
+}
```
elsapet, line 428: 1. Is this sort of map/empty struct pattern okay? It was a Stack Overflow suggestion 😅 
2. Also, does it make sense in go to have something like this as a separate function, or should it all be together in the main `Validate` function?

pkg/util/url/url.go
cfabianski: Not sure I understand the question sorry :(

pkg/util/url/url.go
cfabianski: It felt ok when I reviewed it (for both questions) :)

pkg/util/url/url.go
vjerci: `/util/url/` as it is at the moment looks like a good place to me

elsapet: It's probably me being confused :) 
When classifying detections of all type - URL, dependency, etc - we probably want to check if they are vendored or a potential detector by running the methods `isVendored` and `isPotentialDetector`. Right now these methods are private methods in the URL utils. I'm trying to ask where the best place to put them is, if we want them to be shared methods.

pkg/util/url/url_test.go
```diff
@@ -283,3 +286,312 @@ func TestPrepareURLValue(t *testing.T) {
...
+func TestValidate(t *testing.T) {
+	tests := []struct {
+		Name       string
+		URL        string
+		Data       *report.Detection
+		isInternal bool
+		Want       *url.ValidationResult
+	}{
+		{
+			Name:       "when a detection is from a potential detector",
```
vjerci, line 299: can we not snapshot test this with coupaloy.
What i propose is to have each test case of just `URL`, `Data`, `IsInternal` and snapshot/compare results of it as a whole.

to me having 31 test cases (i suppose it is 100% code coverage) we need to manually input if this function gets feature changed feels wrong.

@cfabianski any thoughts?

vjerci (commented):
I just have a question about tests which is i guess more for @cfabianski .
good job @elsapet  :)

pkg/util/url/url_test.go
cfabianski: I like that way tbh. Feels easy to update/understand than a snapshot. Let’s see how it goes I guess :-)

vjerci: ok i guess thats the standard we should follow :)

pkg/util/url/url.go
cfabianski: I'm fine with what we have now :)

PR #52, opened 2022-10-18
Add func to check domain resolution

elsapet:
## Description
Add `IsReachable` function to URL utility - to be called during interface classification (future PR)

## Checklist

### DIFF
--- a/pkg/util/url/reachable_test.go
+++ b/pkg/util/url/reachable_test.go
@@ -1,0 +1,127 @@
+package url_test
+
+import (
+	"context"
+	"errors"
+	"net"
+	"testing"
+	"time"
+
+	"github.com/bearer/curio/pkg/util/url"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIsReachable(t *testing.T) {
+	tests := []struct {
+		Name, URL      string
+		MockLookUpAddr func(ctx context.Context, addr string) ([]string, error)
+		MockLookUpNS   func(ctx context.Context, name string) ([]*net.NS, error)
+		Want           bool
+	}{
+		{
+			Name: "when we have non-ASCII chars",
+			URL:  "meghívó-måradt.com",
+			Want: true,
+		},
+		{
+			Name: "when the domain contains wildcards and the non wildcard part is not a valid TLD",
+			URL:  "_oidc-client-*-partial.tf",
+			MockLookUpAddr: func(ctx context.Context, addr string) ([]string, error) {
+				return []string{}, nil
+			},
+			Want: false,
+		},
+		{
+			Name: "when the domain is static and resolves",
+			URL:  "www.example.co.za",
+			Want: true,
+		},
+		{
+			Name: "when the domain is static and does not resolve",
+			URL:  "www.example.co.za",
+			MockLookUpAddr: func(ctx context.Context, addr string) ([]string, error) {
+				return []string{}, nil
+			},
+			Want: false,
+		},
+		{
+			Name: "when the domain is static and the DNS server has an error",
+			URL:  "www.example.co.za",
+			MockLookUpAddr: func(ctx context.Context, addr string) ([]string, error) {
+				return []string{}, errors.New("just being nervous")
+			},
+			Want: false,
+		},
+		{
+			Name: "when the domain is static it does not lookup the nameserver",
+			URL:  "www.example.co.za",
+			MockLookUpNS: func(ctx context.Context, name string) ([]*net.NS, error) {
+				panic("we shouldn't reach this")
+			},
+			Want: true,
+		},
+		{
+			Name: "when a wildcard domain does not resolve and its static domain has no nameserver",
+			URL:  "*-test.example.co.uk",
+			MockLookUpAddr: func(ctx context.Context, addr string) ([]string, error) {
+				return []string{}, nil
+			},
+			MockLookUpNS: func(ctx context.Context, name string) ([]*net.NS, error) {
+				return []*net.NS{}, nil
+			},
+			Want: false,
+		},
+		{
+			Name: "when a wildcard domain resolves and its static domain has no nameserver",
+			URL:  "*-test.example.co.uk",
+			MockLookUpNS: func(ctx context.Context, name string) ([]*net.NS, error) {
+				return []*net.NS{}, nil
+			},
+			Want: true,
+		},
+		{
+			Name: "when a wildcard domain does not resolve but its static domain has a nameserver",
+			URL:  "*-test.example.co.uk",
+			MockLookUpAddr: func(ctx context.Context, addr string) ([]string, error) {
+				return []string{}, nil
+			},
+			Want: true,
+		},
+		{
+			Name: "when both the DNS and nameserver lookup raises errors for a wildcard domain",
+			URL:  "*-test.example.co.uk",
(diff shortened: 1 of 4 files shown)


pkg/util/url/url.go
```diff
@@ -479,3 +495,37 @@ func isPotentialDetector(detectorType detectors.Type) bool {
...
 }
+
+func isNameserver(myURL string, staticDomain string) bool {
+	if staticDomain == "" {
+		return false
+	}
+
+	nameserver, err := net.LookupNS(staticDomain)
+	if err != nil {
+		// todo: return true? See Ruby
```
elsapet: In the Ruby app, if e.g. a look up returns an error, we return false for some errors and nil for others. In the nil case, we consider the domain reachable (I think): https://github.com/Bearer/api-management-app/blob/main/app/services/detection_engine/classify_detections/domain_reachable.rb#L96

pkg/util/url/url.go
```diff
@@ -20,6 +22,9 @@ const suffixPattern = "(?:/|\\z)"
 var regexpReplaceMatcher = regexp.MustCompile(`<\w+>`)
 var regexpVariableMatcher = regexp.MustCompile(`\A[*\/.-:]+\z`)
 
+// for domain resolution
+var regexpStaticDomainMatcher = regexp.MustCompile(`(?<=^|\.)[^*]+$`)
```
cfabianski: You are missing a `P` :)
https://stackoverflow.com/questions/30483652/how-to-get-capturing-group-functionality-in-go-regular-expressions

cfabianski (commented):
All the other questions have been answered in a Zoom call :)

pkg/util/url/url.go
elsapet: From Zoom: since we're not caching, we'll return `false` for any errors encountered. In Ruby app, the `nil` cases were for caching purposes.

pkg/util/url/url_test.go
```diff
@@ -11,6 +14,142 @@ import (
...
+			Want:         false,
+		},
+		// todo: public suffix validation not available for go lib
+		// {
+		// 	Name:           "when PublicSuffix does not recognise the domain",
+		// 	URL:            "kobe.jp",
+		// 	MockLookUpAddr: defaultMockLookUpAddr,
+		// 	MockLookUpNS:   defaultMockLookUpNS,
+		// 	Want:           false,
+		// },
```
elsapet: @cfabianski - this case is difficult to catch without `PublicSuffix.valid?` which is not available in the Go library. 
Myself I think we should be okay to leave this out and rely on the other validation checks we perform during classification - any thoughts your side?

cfabianski: I agree 👍

elsapet: Coolio, then I shall remove this test 🙏

pkg/util/url/url.go
```diff
@@ -189,6 +194,23 @@ type ValidationResult struct {
 	Reason string
 }
 
+// given a valid URL, is it reachable?
+func IsReachable(myURL string, timeout time.Duration) bool {
```
vjerci: this file is getting quite big perhaps it would be wise to split it into 2 files,
there is nothing wrong with having `reachable.go` in `pkg/util/url/reachable.go`

pkg/util/url/url_test.go
```diff
@@ -11,6 +14,133 @@ import (
...
+
+	timeDuration := 3 * time.Second
+
+	originalLookUpAddr := url.LookUpAddr
+	originalLookUpNS := url.LookUpNS
+
+	for _, testCase := range tests {
+		t.Run(testCase.Name, func(t *testing.T) {
+			// add mocking
+			url.LookUpAddr = testCase.MockLookUpAddr
```
vjerci: this is bad you are mutating a package here causing subsequent tests to have unexpected behaviour.

Consider this example: subsequent test `TestSomething2` will fail.
---
```
package something

var Something bool = true
```

```
package something_test

import (
	"testing"

	"github.com/bearer/curio/something"
	"github.com/stretchr/testify/assert"
)

func TestSomething(t *testing.T) {
	something.Something = false
	assert.Equal(t, something.Something, false)
}

func TestSomething2(t *testing.T) {
	assert.Equal(t, something.Something, true)
}
```
Possible solutions:
---
use cleanup to reset defaults
---
This one is in my opinion bad as you are still mutating package.
I would treat exposed package vars as global variables and mutating them always feels bad.
I see you are using something similar with remove mocking
```
var defaultVarValue = something.Something

func TestSomething(t *testing.T) {
	t.Cleanup(resetDefaults)
	something.Something = false
	assert.Equal(t, something.Something, false)
}

func TestSomething2(t *testing.T) {
	t.Cleanup(resetDefaults)
	assert.Equal(t, something.Something, true)
}

func resetDefaults() {
	something.Something = defaultVarValue
}
```
Structure code so it is easier to override/test
---
```
package something

type Holder struct {
	Something bool
}

func NewHolder() *Holder {
	return &Holder{
		Something: true,
	}
}
```
```
package something_test

import (
	"testing"

	"github.com/bearer/curio/something"
	"github.com/stretchr/testify/assert"
)

func TestSomething(t *testing.T) {
	holder := something.NewHolder()
	holder.Something = false
	assert.Equal(t, holder.Something, false)
}

func TestSomething2(t *testing.T) {
	holder := something.NewHolder()
	assert.Equal(t, holder.Something, true)
}

```
if you go this way you will probabbly have something like this
```
type Reachable struct {
	LookUpAddr func(ctx context.Context, addr string) ([]string, error)
	LookUpNS   func(ctx context.Context, addr string) ([]*net.NS, error)
}

func NewReachable() *Reachable {
	var resolver = net.Resolver{PreferGo: true}

	return &Reachable{
		LookUpAddr: resolver.LookupAddr,
		LookUpNS:   resolver.LookupNS,
	}
}

func (reachable *Reachable) CanReach() {
	// ....
	reachable.LookUpAddr(context.Background(), addr)
}
```

elsapet: Thanks for the guidance @vjerci 🙏 I'll go the preferred restructuring route 👍

pkg/util/url/reachable_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	for _, testCase := range tests {
+		// reset mocks
+		reachable := url.NewReachable()
+		reachable.LookUpAddr = func(ctx context.Context, addr string) ([]string, error) {
+			return []string{"www.example.co.za"}, nil
+		}
+		reachable.LookUpNS = func(ctx context.Context, name string) ([]*net.NS, error) {
+			return []*net.NS{{Host: "example.co.za"}}, nil
+		}
```
elsapet: @vjerci is it okay to keep the `for` loop here, and create a new `reachable` instance in each iteration, or do we prefer separate tests entirely?

elsapet:
@vjerci thanks again for your feedback. I've had another go - let me know how it's looking to you 🙏

pkg/util/url/reachable_test.go
vjerci: usually its inside t.Run but in my honest opinion it doesn't matter much

vjerci (dismissed):
looks great :)

vjerci (approved):

PR #57, opened 2022-10-19
Debug Flag

vjerci:
## Description
Add support for debug flag `--debug` to enable debug, by default it is false and sets up log level to `info` for both processing worker and main process.
It also adds a default logger as an util for writing clean logs to cli (the ones without timestamp and that are not json serliazed).

This pr doesn't handle passing down all configuration flags and options to processing worker.
If the need for it arises it should be handled in separate ticket.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ curio scan ./curio-ci-test/Pipfile.lock
 ### Scan Flags
 
 - `--skip-path` specify the comma separated files and directories to skip (supports \* syntax), eg. --skip-path users/\*.go,users/admin.sql
-- `--debug` enable debug logs (default 0)
+- `--debug` enable debug logs
 
 ### Worker Flags
 
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ curio scan ./curio-ci-test/Pipfile.lock
 ### Scan Flags
 
 - `--skip-path` specify the comma separated files and directories to skip (supports \* syntax), eg. --skip-path users/\*.go,users/admin.sql
+- `--debug` enable debug logs (default 0)
 
 ### Worker Flags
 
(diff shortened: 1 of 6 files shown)


cfabianski (changes requested):
I was thinking more about using `--debug` and not having to specify a value at all. 

so

```bash
curio scan <project> --debug
```

and not

```
curio scan <project> --debug=<something>
```

In the current state, that doesn't feel right.

vjerci:
> I was thinking more about using `--debug` and not having to specify a value at all.
> 
> so
> 
> ```shell
> curio scan <project> --debug
> ```
> 
> and not
> 
> ```
> curio scan <project> --debug=<something>
> ```
> 
> In the current state, that doesn't feel right.

I've tested it and it actually works both ways.
To enable debug you can do `curio scan <project> --debug` or `curio scan <project> --debug=1`

README.md
```diff
@@ -65,6 +65,7 @@ curio scan ./curio-ci-test/Pipfile.lock
 ### Scan Flags
 
 - `--skip-path` specify the comma separated files and directories to skip (supports \* syntax), eg. --skip-path users/\*.go,users/admin.sql
+- `--debug` enable debug logs (default 0)
```
cfabianski:
```suggestion
- `--debug` enable debug logs
```

pkg/commands/process/balancer/process.go
```diff
@@ -44,8 +44,13 @@ func (process *Process) StartProcess(task *workertype.ProcessRequest) error {
...
 	}
 
+	debugArgument := "0"
+	if process.config.Scan.Debug {
+		debugArgument = "1"
+	}
+
 	log.Debug().Msgf("spawning on port %s", port)
-	cmd := exec.Command(currentCommand, "processing-worker", "--port", port)
+	cmd := exec.Command(currentCommand, "processing-worker", "--port="+port, "--debug="+debugArgument)
```
cfabianski:
```
cmd := exec.Command(currentCommand, "processing-worker", "--port="+port, "--debug")
```
or this
```
cmd := exec.Command(currentCommand, "processing-worker", "--port="+port)
```

cfabianski (approved):

PR #58, opened 2022-10-20
Check domain resolution during classification

elsapet:
## Description
* Some refactoring to reachable util (renamed domainResolver) 
* Add domain checking to url util and interface classifier

## Checklist

### DIFF
--- a/pkg/classification/interfaces/interfaces.go
+++ b/pkg/classification/interfaces/interfaces.go
@@ -31,11 +31,13 @@ type Classification struct {
 type Classifier struct {
 	Recipes                []Recipe
 	InternalDomainMatchers []*regexp.Regexp
+	DomainResolver         *url.DomainResolver
 }
 
 type Config struct {
 	Recipes                []db.Recipe
 	InternalDomainMatchers []*regexp.Regexp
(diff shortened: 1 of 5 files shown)


pkg/classification/interfaces/interfaces.go
```diff
@@ -79,14 +81,19 @@ func New(config Config) (*Classifier, error) {
 		preparedRecipes = append(preparedRecipes, preparedRecipe)
 	}
 
-	return &Classifier{Recipes: preparedRecipes, InternalDomainMatchers: config.InternalDomainMatchers}, nil
+	return &Classifier{
+		Recipes:                preparedRecipes,
+		InternalDomainMatchers: config.InternalDomainMatchers,
+		DomainResolver:         config.DomainResolver,
```
elsapet, line 87: I added domain resolver at this level (rather than e.g. passing the individual options like timeout and enabled/disabled) because we should be able to share one domain resolver for all of the URLs that we are classifying

cfabianski (dismissed):
👏

cfabianski (approved):

vjerci (approved):

PR #59, opened 2022-10-20
Support version command

vjerci:
## Description
This pr adds support for version command as well as hardcoding tag version in binaries.

## Checklist

### DIFF
--- a/goreleaser/linux.yaml
+++ b/goreleaser/linux.yaml
@@ -15,7 +15,8 @@ builds:
       - amd64
     ldflags:
       - -s -w
-      - -X github.com/bearer/curio/cmd/curio/build.Version={{.Version}}
+      - -X "github.com/bearer/curio/cmd/curio/build.Version={{.Version}}"
+      - -X "github.com/bearer/curio/cmd/curio/build.CommitSHA={{.Commit}}"
 snapshot:
   name_template: "{{ incpatch .Version }}-next"
 changelog:
--- a/goreleaser/linux.yaml
+++ b/goreleaser/linux.yaml
@@ -14,6 +14,7 @@ builds:
       - 386
       - amd64
     ldflags:
+      - -s -w
       - -X github.com/bearer/curio/cmd/curio/build.Version={{.Version}}
 snapshot:
   name_template: "{{ incpatch .Version }}-next"
--- a/goreleaser/linux.yaml
+++ b/goreleaser/linux.yaml
@@ -13,6 +13,8 @@ builds:
     goarch:
       - 386
       - amd64
+    ldflags:
+      - -X github.com/bearer/curio/cmd/curio.Version={{.Version}}
 snapshot:
   name_template: "{{ incpatch .Version }}-next"
 changelog:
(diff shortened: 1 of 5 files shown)


goreleaser/linux.yaml
```diff
@@ -13,6 +13,8 @@ builds:
     goarch:
       - 386
       - amd64
+    ldflags:
```
cfabianski, line 16: Could you explain what this does please?
Just checking that the version would be the version defined in the tag and/or the commit SHA when we release

pkg/commands/app.go
```diff
@@ -204,41 +203,23 @@ func NewConfigCommand() *cobra.Command {
...
+func NewVersionCommand(version string) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "version [flags]",
+		Use:   "version",
 		Short: "Print the version",
 		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			showVersion(versionFormat, cmd.Version, outputWriter)
-
+			output.DefaultLogger().Msgf("curio version: %s", version)
```
cfabianski: Apparently you forgot something here :)

vjerci: yeah am waiting for debug pr to be approved so that i can use it.

cfabianski: wait no more, it's been approved ;)

goreleaser/linux.yaml
vjerci: this sets golang variable  called `Version` in package `github.com/bearer/curio/cmd/curio` to a github tag.
So once it is built it should have the correct `Version` embedded inside the program.

cfabianski: Could we have both a SHA **and** a Tag when available? 🙏

pkg/commands/app.go
vjerci: sry was paring with with elizabeth for some great code features  :)

goreleaser/linux.yaml
vjerci: sure

vjerci: should the version command output it
```
curio version: {tag}
sha: {SHA}
```

cfabianski: Yes that's perfect IMO cc @gotbadger @markmichon

pkg/commands/app.go
```diff
@@ -215,41 +214,23 @@ func NewConfigCommand() *cobra.Command {
...
 			return nil
 		},
-		SilenceErrors: true,
-		SilenceUsage:  true,
+		SilenceErrors: false,
+		SilenceUsage:  false,
```
cfabianski: Is that intended?

cfabianski: https://github.com/spf13/cobra/blob/860791844ed3a2e544a9b9bbbcb14144a948ad20/command.go#L222-L226

vjerci: nice catch, no its not needed its the default

goreleaser/linux.yaml
vjerci: implemented and resolving this one for now, if its a problematic for some reason we can always fix it later.

cfabianski (approved):

PR #61, opened 2022-10-20
feat: implement config serialization and loading between main and processing worker

elsapet:
## Description
Serialize config options and pass these from main to processing worker
Scan now has access to config with all flags

## Checklist

### DIFF
--- a/pkg/commands/process/worker/worker.go
+++ b/pkg/commands/process/worker/worker.go
@@ -6,14 +6,22 @@ import (
 	"net/http"
 	"runtime"
 
+	classsification "github.com/bearer/curio/pkg/classification"
+	config "github.com/bearer/curio/pkg/commands/process/settings"
 	"github.com/bearer/curio/pkg/commands/process/worker/blamer"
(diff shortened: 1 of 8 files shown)


vjerci (approved):

pkg/commands/process/worker/worker.go
```diff
@@ -6,14 +6,22 @@ import (
...
 	"github.com/bearer/curio/pkg/commands/process/worker/work"
 	"github.com/bearer/curio/pkg/detectors"
 	"github.com/bearer/curio/pkg/scanner"
 )
 
-func Start(port string) error {
-	err := http.ListenAndServe(`localhost`+port, http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+func Start(port string, config config.Config) error {
+	classifier := classsification.NewClassifier(&classsification.Config{Config: config})
+	err := detectors.SetupCustomDetector(config.CustomDetector.RulesConfig)
```
cfabianski, line 19: Would probably be better to rename `SetupCustomDetector` to `InitCustomDetector` maybe?

PR #68, opened 2022-10-21
Add internal domains flag

elsapet:
## Description
Add internal domains flag and use this during classifier instantiation

We accept internal domains as strings, and then parse these as regular expressions when initializing the interfaces classifier

## Checklist

### DIFF
--- a/docs/reference/commands.md
+++ b/docs/reference/commands.md
@@ -37,7 +37,7 @@ curio scan [FLAGS] [PATH]
 - `--debug` enable debug logs
 - `--disable-domain-resolution` skip attempt to resolve detected domains during classification (default false)
 - `--domain-resolution-timeout` set timeout when attempting to resolve detected domains during classification
-- `--internal-domains` define regular expressions for better classification of private or unreachable domains eg. --internal-domains="\*.my-company.com\,\private.sh\"
+- `--internal-domains` define regular expressions for better classification of private or unreachable domains eg. --internal-domains="*.my-company.com,private.sh"
 
 #### Worker Flags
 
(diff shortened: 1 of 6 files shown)


docs/reference/commands.md
```diff
@@ -37,6 +37,7 @@ curio scan [FLAGS] [PATH]
 - `--debug` enable debug logs
 - `--disable-domain-resolution` skip attempt to resolve detected domains during classification (default false)
 - `--domain-resolution-timeout` set timeout when attempting to resolve detected domains during classification
+- `--internal-domains` define regular expressions for better classification of private or unreachable domains eg. --internal-domains="\*.my-company.com\,\private.sh\"
```
cfabianski: Not sure I like the format with the slashes. Should be simpler to write IMO

cfabianski:
```suggestion
- `--internal-domains` define regular expressions for better classification of private or unreachable domains eg. --internal-domains="*.my-company.com,private.sh"
```

It should be this IMO

cfabianski: Also, should it be `--safelist-domains` instead?

elsapet: Turns out I was confused and the `\` is not expected after all 🙈  I'll update the docs 🙏

elsapet: No strong opinions on the naming my side, but if we do change to `--safelist-domains`, for coherence we should also update the decision reasons (e.g. update `"internal_domain_path_contains_api_or_auth"` to `"safelist_domain_path_contains_api_or_auth"`), and I'd prefer to make these changes in a separate PR 🙏

cfabianski (approved):

PR #72, opened 2022-10-24
feat(classification): add data type classification data to db.go

elsapet:
## Description
* Add data types and related data type patterns to default database
* Refactor db.Default() usage since it's now more than just recipes

## Checklist


pkg/classification/db/db.go
```diff
@@ -28,7 +46,44 @@ type RecipeType string
...
+func Default() DefaultDB {
+	return DefaultDB{
+		Recipes:                        defaultRecipes(),
+		DataTypes:                      defaultDataTypes(),
+		DataTypeClassificationPatterns: defaultDataTypeClassificationPatterns(),
+		KnownPersonObjectPatterns:      defaultKnownPersonObjectPatterns(),
+	}
+}
+
+func defaultRecipes() []Recipe {
```
elsapet, line 85: I'm not sure if there's a golang way to extract the code shared between these `defaultX()` methods? It's all the same except for the types 🙈

pkg/classification/db/db.go
```diff
@@ -43,27 +98,99 @@ func Default() []Recipe {
...
 	if err != nil {
-		log.Fatalf("failed to unmarshal db %e", err)
-		return nil, err
+		log.Fatalln(err)
+	}
+
+	for _, file := range files {
+		val, err := knownPersonObjectPatternsDir.ReadFile("known_person_object_patterns/" + file.Name())
+		if err != nil {
+			fmt.Println(err)
```
cfabianski: Should probably be `log.Errorf`

pkg/classification/db/db.go
```diff
@@ -43,27 +98,99 @@ func Default() []Recipe {
...
+
+	files, err := dataTypeClassificationPatternsDir.ReadDir("data_type_classification_patterns")
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	for _, file := range files {
+		val, err := dataTypeClassificationPatternsDir.ReadFile("data_type_classification_patterns/" + file.Name())
+		if err != nil {
+			fmt.Println(err)
```
cfabianski: Should probably be `log.Errorf`

cfabianski (commented):
Let's be careful with the logging. 
Looks good otherwise.

cfabianski (approved):

pkg/classification/db/db.go
elsapet: Resolving this comment so I can merge but still open for advice on improving this

PR #74, opened 2022-10-24
fix(classification): prefer LookupIPAddr to LookupAddr

elsapet:
## Description
We need to use LookupIPAddr and not LookupAddr - this fixes the issue of e.g. https://datadog.eu being classed as unreachable

Also some renaming clean-up since we are in the area

DD report after this fix: 

```
{
		"classification": {
			"decision": {
				"reason": "no_subdomain",
				"state": "potential"
			},
			"recipe_match": false,
			"url": "https://datadoghq.eu"
		},
		"commit_sha": "",
		"detector_type": "ruby",
		"source": {
			"column_number": 6,
			"filename": ".",
			"language": "Ruby",
			"language_type": "programming",
			"line_number": 4,
			"text": "\"https://datadoghq.eu\""
		},
		"type": "interface_classified",
		"value": {
			"type": "url",
			"value": {
				"parts": [
					{
						"type": "string",
						"value": "https://datadoghq.eu"
					}
				]
			}
		}
	}
```

## Checklist

### DIFF
--- a/pkg/util/url/domain_resolver.go
+++ b/pkg/util/url/domain_resolver.go
@@ -15,8 +15,8 @@ var regexpDomainSplitMatcher = regexp.MustCompile(`\*\s*(.*?)\s*\.`)
 type DomainResolver struct {
 	Enabled      bool
 	Timeout      time.Duration
-	LookupIPAddr func(ctx context.Context, addr string) ([]net.IPAddr, error)
-	LookupNS     func(ctx context.Context, addr string) ([]*net.NS, error)
+	LookupIPAddr func(ctx context.Context, host string) ([]net.IPAddr, error)
+	LookupNS     func(ctx context.Context, host string) ([]*net.NS, error)
 }
 
 func NewDomainResolver(enabled bool, timeout time.Duration) *DomainResolver {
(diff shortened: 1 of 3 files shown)


cfabianski:
Could you explain the difference between `lookupIpAddr` and `lookupAddr` please? Why not `lookupHost`?

elsapet:
@cfabianski thanks for the question. [From the docs](https://pkg.go.dev/net#Resolver), `lookupAddr` "performs a reverse lookup for the given address". Because we were giving it a host and not an address, it was failing to resolve when expected.

I don't think it's a big difference if we use `lookupIPAddr` or `lookupHost` - they both take a host and not an address and they both use the local resolver. My reason for choosing `lookupIPAddr` is because it is the closer option to what we have in the Rails app (`getaddress`, which [returns an IP address given a host](https://ruby-doc.org/stdlib-2.5.1/libdoc/resolv/rdoc/Resolv.html#method-c-getaddress))

cfabianski:
> I don't think it's a big difference if we use `lookupIPAddr` or `lookupHost` - they both take a host and not an address and they both use the local resolver. My reason for choosing `lookupIPAddr` is because it is the closer option to what we have in the Rails app (`getaddress`, which [returns an IP address given a host](https://ruby-doc.org/stdlib-2.5.1/libdoc/resolv/rdoc/Resolv.html#method-c-getaddress))

Fair enough. Thanks for the explanation!

cfabianski (dismissed):

cfabianski (approved):

PR #81, opened 2022-10-26
test: Add first integration tests

spdawson:
## Description

This adds snapshot-based integration testing for the following command and option flag combinations.

* `curio help`
* `curio version`

## Checklist

### DIFF
--- a/integration/flags/.snapshots/TestMetadataFlags-version
+++ b/integration/flags/.snapshots/TestMetadataFlags-version
@@ -1,0 +1,5 @@
+
+--
+curio version: dev
+sha: devSHA
+
--- a/integration/internal/testhelper/testhelper.go
+++ b/integration/internal/testhelper/testhelper.go
@@ -1,0 +1,130 @@
+package testhelper
+
+import (
+	"bytes"
+	"io"
+	"os"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/bearer/curio/cmd/curio/build"
+	"github.com/bearer/curio/pkg/commands"
+	"github.com/bearer/curio/pkg/commands/process/balancer"
+	"github.com/bradleyjkemp/cupaloy"
+	"github.com/rs/zerolog/log"
+)
+
+type TestCase struct {
+	Name          string
+	Arguments     []string
+	ShouldSucceed bool
+}
+
+func NewTestCase(name string, arguments []string) *TestCase {
+	return &TestCase{
+		Name: name,
+		Arguments: arguments,
+		ShouldSucceed: true,
+	}
+}
+
+func executeApp(arguments []string, port int) (string, error) {
+	app := commands.NewApp(build.Version, build.CommitSHA)
+
+	if arguments[0] == "scan" {
+		arguments = append(arguments, "--existing-worker=http://localhost:"+strconv.Itoa(port))
+	}
+
+	stdoutReader, stdoutWriter, err := os.Pipe()
+	if err != nil {
+		panic(err)
+	}
+
+	stderrReader, stderrWriter, err := os.Pipe()
+	if err != nil {
+		panic(err)
+	}
+
+	stdout := os.Stdout
+	os.Stdout = stdoutWriter
+	defer func() {
+		os.Stdout = stdout
+	}()
+
+	stderr := os.Stderr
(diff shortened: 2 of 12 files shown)


integration/flags/.snapshots/TestMetadataFlags-version
```diff
@@ -0,0 +1 @@
+
```
cfabianski, line 5: Why is it empty?

spdawson: Currently unknown; this is not yet in a working state. The other problem is that the tests for the `--format`, `--report` amd `--output` flags are hanging.

integration/flags/report_flags_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/bearer/curio/integration/internal/testhelper"
+)
+
+func TestReportFlags(t *testing.T) {
+	tests := []testhelper.TestCase{
+		{
+			Name: "format",
+			Arguments: []string{"scan", filepath.Join("testdata", "simple"), "--format", "json", "--config-file", filepath.Join("testdata", "config.yml")},
+		},
```
cfabianski: It doesn't seem like we have the output of those tests anywhere, is that expected?

cfabianski (commented):
I think it's a good draft! Would there be any way to check the arguments used by the program?

integration/flags/report_flags_test.go
spdawson: :sweat_smile: That's right: these tests are currently not working &mdash; they just hang the test runner. I haven't worked out what's going wrong yet.

integration/internal/testhelper/testhelper.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"bytes"
+	"io/ioutil"
+	"testing"
+	"github.com/bradleyjkemp/cupaloy"
+	"github.com/bearer/curio/cmd/curio/build"
+	"github.com/bearer/curio/pkg/commands"
+)
+
+type TestCase struct {
```
vjerci, line 18: I don't think we need this file to be in separate package.
usually test type and running instructions are defined in same function.

This is standard practice:
https://github.com/aquasecurity/trivy/blob/main/integration/docker_engine_test.go#L20

integration/internal/testhelper/testhelper.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func RunTests(t *testing.T, tests []TestCase) {
+	for _, test := range tests {
+		t.Run(test.Name, func(t *testing.T) {
+			app := commands.NewApp(build.Version, build.CommitSHA)
+
+			out_buffer := bytes.NewBufferString("")
+			err_buffer := bytes.NewBufferString("")
+			app.SetOut(out_buffer)
```
vjerci: I think that it might be easiest to replace os.StdOut and os.StdErr with our own writer, instead of setting app output.

integration/internal/testhelper/testhelper.go
```diff
@@ -0,0 +1,130 @@
...
+	go func() {
+		err := startWorker(port)
+		if err != nil {
+			log.Fatal().Msgf("failed to start worker: %s", err)
+		}
+	}()
+
+	// this needs to be here since otherwise viper is getting written twice concurrently from 2 gorutines
+	// we need to find a way to let main program know viper has finished loading config
+	time.Sleep(1 * time.Second)
```
vjerci, line 113: can we perhaps increase this timeout to 5seconds? 
I would rather wait a bit than risk intermittent failures

vjerci (dismissed):
Good job 💪 🎊

integration/internal/testhelper/testhelper.go
spdawson: I've done some local testing with this timeout set to very short values (e.g. as low as 1ms), and I haven't seen any problems. I wonder if we can leave this as-is for the time being, and increase it if we do find it needs to be longer?

My reasoning is that I don't want to introduce a 5s delay into every integration test run: we will currently incur this delay for every test file that invokes `RunTests`. Even this current 1s delay could ultimately prove to be unacceptably long... :sweat_smile:

vjerci (approved):

PR #82, opened 2022-10-26
feat: integrate schema classification

vjerci:
## Description
this pr refractores detections so that they live inside `/pkg/report/detections`
It also adds support for generic `DatatTypeable` interface handling, allowing helper functions to support different datatype structures.
It integrates datatypes into classification flow

## Checklist


pkg/report/schema/datatype/datatype.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/bearer/curio/pkg/report/detections"
+	"github.com/bearer/curio/pkg/report/detectors"
+	"github.com/bearer/curio/pkg/report/schema"
+)
+
+type ReportDataType interface {
+	AddDataType(detectionType detections.DetectionType, detectorType detectors.Type, generator nodeid.Generator, values map[parser.NodeID]*DataType)
+}
+
+type DataType struct {
```
vjerci, line 18: @elsapet I've changed a bit how datatypes work, I think this file is important for you.

pkg/classification/schema/schema_test.go
```diff
@@ -42,22 +42,22 @@ func TestSchema(t *testing.T) {
...
 					UUID: "1",
-				},
-				Classification: schema.Classification{
-					Name: "personal data",
-				},
-				Properties: map[string]schema.ClassifiedDatatype{
-					"address": {
-						Classification: schema.Classification{
-							Name: "personal data",
+					Properties: map[string]datatype.DataTypable{
```
vjerci, line 45: @elsapet This file has also been changed a bit

pkg/report/schema/datatype/datatype.go
```diff
@@ -0,0 +1,172 @@
...
+	Node       *parser.Node
+	Name       string
+	Type       string
```
cfabianski, line 21: Could you add a `SimpleType` here too so that we can pass it to the classification engine?

pkg/report/schema/datatype/datatype.go
```diff
@@ -0,0 +1,172 @@
...
+	TextType   string
+	Properties map[string]DataTypable
+	IsHelper   bool // helper dataTypes and their child datatypes don't get exported
```
cfabianski, line 24: Not sure I understand that.

cfabianski (commented):
A lot of incidental changes. Left 2 comments but looks good in general.
@elsapet I'll let you approve if you are okay with that

pkg/report/schema/datatype/datatype.go
vjerci: It is just something I've copied from previous workings.
it is something we needed for languages datatypes, If you mark it as helper it doesn't get exported.
As far as i recall it is used for taking care of duplicates, and I'm not sure if simply deleting it from parent properties was doable.

pkg/report/schema/datatype/datatype.go
vjerci: Naming is a bit wierd here but in my opinion fixing it is outside of scope of this pr.
`Type` is actually simpletype while `TextType` is full type
So `type User struct` would have `type` of `object` and `TextType` of `User` .

cfabianski: What is name then here?

elsapet (approved):

pkg/report/schema/datatype/datatype.go
vjerci: i think name is also `user`

``` 
type User struct {
    id int
    address string
}
```

from this data we should have 
``` 
DataType{
	Name: "User",
	Type: schema.SimpleTypeObject,
	TextType: "User",
	Properties: map[string]DataTypable{
		"id": &DataType{
			Name: "id",
			Type: schema.SimpleTypeNumber,
			TextType: "int",
		},
		"address": &DataType{
			Name: "address",
			Type: schema.SimpleTypeString,
			TextType: "string",
		},
	},
} 
```

PR #83, opened 2022-10-26
fix: config erroring when config is not found

vjerci:
## Description
We have probably upgraded viper version, making previous error checking mehanics for config existance invalid.
This PR fixes it.

## Checklist

### DIFF
--- a/pkg/commands/scan.go
+++ b/pkg/commands/scan.go
@@ -1,9 +1,7 @@
 package commands
 
 import (
-	"errors"
 	"fmt"
-	"os"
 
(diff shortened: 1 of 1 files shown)


cfabianski (approved):

cfabianski:
Should we have a test for it? I guess it would also be picked up by @spdawson's work regarding integration testing

vjerci:
> Should we have a test for it? I guess it would also be picked up by @spdawson's work regarding integration testing

I don't know how to test it besides integration.

PR #85, opened 2022-10-26
feat(classification): classify known object schema detections

elsapet:
## Description
* Add classification for "known person object" schema detections (top row of [this Miro](https://miro.com/app/board/uXjVOZ9xlro=/))
* Ensure all data type classification files are added (friendly name is not unique so we include ID when building filename)
* Some simplification of classify methods (remove detector type checks and cloud related things)

Note: we're assuming only 1 "layer" of data type nesting (e.g. User object with Name field and not User object with Address object with Postcode field)

Some TODOs / open questions (see comments)

Things to handle in later PRs: 
* Context check (see Miro - example, Health context)
* Handling further nesting (e.g. User -> Address -> Postcode example)
* The other branches in the Miro (Unknown Objects, etc)

## Checklist


pkg/classification/schema/schema.go
```diff
@@ -31,27 +67,240 @@ type DataTypeDetection struct {
...
+
+	// stop_word
+	if objectStopWordDetected(normalizedObjectName) {
+		classifiedDataType.Classification.Decision = classify.ClassificationDecision{
+			State:  classify.Invalid,
+			Reason: "stop_word",
+		}
+	}
+
+	// todo: do we need these checks too? app/services/detection_engine/classify_detections/classify_schema_object_detection.rb
```
elsapet: @cfabianski not sure if we need to include the from `ClassifySchemaObjectDetection` here: https://github.com/Bearer/api-management-app/blob/main/app/services/detection_engine/classify_detections/classify_schema_object_detection.rb#L16-L24

pkg/classification/schema/schema.go
```diff
@@ -31,27 +67,240 @@ type DataTypeDetection struct {
...
+						},
 					},
-					DataType: &datatype.DataType{
-						UUID: "2",
+				}
+
+				continue
+			}
+
+			// todo: check for field stop word?
```
elsapet: @cfabianski both objects and fields can have stop words (if I'm understanding the Rails code). Do we want to include a check for field stop words here? It's not in the Miro but neither are some of the other Object level checks

pkg/util/classify/classify_test.go
```diff
@@ -0,0 +1 @@
+package classify_test
```
elsapet, line 1: Will add some 🙈

pkg/classification/schema/schema.go
elsapet: ^ We don't include them here. These checks are run for unknown objects as a fallback - i.e. the "Match object in known data objects Patterns" card in the Miro

pkg/classification/schema/schema.go
elsapet: ^ Yes, we want to check for a stop word for fields / properties and return early if this is the case

cfabianski (approved):
That seems correct to me! Good job

PR #90, opened 2022-11-01
docs: add docs site 

markmichon:
## Description

Adds an initial version of the docs site

## Checklist


markmichon:
@gotbadger Could you take a peek at the workflow file? 
Few notes:
- Goal is to setup/run node deps on the `docs` directory. `working-directory` should handle that.
- Runs the build command in the package.json
- Uses [this action](https://github.com/marketplace/actions/github-pages-action) to handle copying the build folder `_site` over to the gh-pages branch. This will require you to configure gh-pages in the repo settings.

Unknowns:
- I'm not certain if the `./_site` setting on the final line will reference the working directory or the repo as a whole.
- I've mostly used the action's default config, but we may wish to make builds manual instead of on main push, or release tags + manual? Feel free to update. 

I'll add some no-index settings to the theme before we push this so testing can be done without google crawling.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,41 @@
+name: Build Documentation
+
+on:
+  push:
```
gotbadger: lets just use `workflow_dispatch` till we work out whats what :)

gotbadger (dismissed):

gotbadger (approved):

PR #93, opened 2022-11-01
fix: correct handling of ./ scan target prefix

spdawson:
## Description

When the scan target is specified with a `./` prefix, we currently incorrectly build relative file paths within the project directory.

Also ensure that we correctly handle the cases where a trailing slash is either present or absent, on a directory path specified as scan target.

## Checklist


gotbadger (dismissed):
Thanks for the fix on this one @spdawson

elsapet (dismissed):

elsapet (approved):

PR #97, opened 2022-11-02
docs: resolve live site path issue

markmichon:
## Description

Resolves the issue where relative paths all reference bearer.github.io rather than bearer.github.io/curio.

Includes an upgrade to the pre-release version of 11ty which makes this process less manual.

## Checklist

### DIFF
--- a/docs/.eleventy.js
+++ b/docs/.eleventy.js
@@ -1,27 +1,34 @@
-const syntaxHighlight = require("@11ty/eleventy-plugin-syntaxhighlight")
-const yaml = require("js-yaml")
-const markdownIt = require("markdown-it")
-const markdownItEmoji = require("markdown-it-emoji")
-const now = String(Date.now())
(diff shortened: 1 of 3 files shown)


docs/.eleventy.js
```diff
@@ -1,27 +1,34 @@
...
   eleventyConfig.addShortcode("version", function () {
-    return now
-  })
+    return now;
+  });
 
-  eleventyConfig.addPlugin(syntaxHighlight)
-  eleventyConfig.setLibrary("md", markdownIt({html: true}).use(markdownItEmoji))
+  eleventyConfig.addPlugin(EleventyHtmlBasePlugin, {
+    baseHref: process.env.ELEVENTY_PRODUCTION ? "/curio/" : "/",
```
gotbadger, line 20: do we need to pass this in during build?

markmichon: It's already set in npm build command

gotbadger (approved):

PR #100, opened 2022-11-04
fix(classification): include shared validation checks in dependency classification

elsapet:
## Description
We were missing these checks for dependency detections
I also add a classification decision 

## Checklist

### DIFF
--- a/pkg/classification/dependencies/dependencies.go
+++ b/pkg/classification/dependencies/dependencies.go
@@ -6,6 +6,7 @@ import (
 	"github.com/bearer/curio/pkg/classification/db"
 	"github.com/bearer/curio/pkg/report/dependencies"
 	"github.com/bearer/curio/pkg/report/detections"
+	"github.com/bearer/curio/pkg/util/classify"
 )
 
 type ClassifiedDependency struct {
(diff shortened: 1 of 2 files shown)


pkg/classification/dependencies/dependencies.go
```diff
@@ -45,12 +47,40 @@ func (classifier *Classifier) Classify(data detections.Detection) (*ClassifiedDe
 		return nil, errors.New("detection is not an dependency")
 	}
 
+	if classify.IsVendored(data.Source.Filename) {
+		return &ClassifiedDependency{
+			Detection: &data,
+			Classification: &Classification{
+				Decision: classify.ClassificationDecision{
+					State:  classify.Invalid,
+					Reason: classify.IncludedInVendorFolderReason,
```
elsapet: Question: if it's a valid detection but there's no recipe match we return `nil` - would we prefer to do this for e.g. invalid detections in this case?

cfabianski (approved):

pkg/classification/dependencies/dependencies.go
elsapet: ^ We return `nil` in all cases for consistency. We can add Invalid classification states and reasons if/when they are requested.

PR #105, opened 2022-11-09
chore: Consolidate output and error stream usage

spdawson:
## Description

Curio's CLI architecture is obtained from the use of the Cobra framework. If we use the Cobra output and error streams correctly, then it would not be necessary to temporarily switch stdout/stderr in integration tests.

This change makes a start on consolidating and standardizing the various output stream mechanisms that are currently in use. The primary objective is to be capture the Cobra output and error streams directly in the integration tests. There may be further work to do around consolidating/standardizing output and error stream usage in the codebase.

## Checklist

### DIFF
--- a/integration/flags/.snapshots/TestMetadataFlags-version
+++ b/integration/flags/.snapshots/TestMetadataFlags-version
@@ -1,5 +1,5 @@
-
---
 curio version: dev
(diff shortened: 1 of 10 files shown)


cfabianski (approved):
Good job!

PR #111, opened 2022-11-10
test: Hide the progress bar when running integration tests

spdawson:
## Description

Add a `--quiet` flag to suppress non-essential output. In particular, this hides the progress bar, which is problematical for the integration tests.

The progress bar itself is not problematical, but the iterations per second ("files/s") component causes problems for snapshot-based testing, as the processing rate will vary between runs.

After this change, the progress bar is shown unless either (or both) of the `--debug` or `-quiet` flags are provided.

## Checklist

### DIFF
--- a/integration/internal/testhelper/testhelper.go
+++ b/integration/internal/testhelper/testhelper.go
@@ -34,7 +34,7 @@ func executeApp(arguments []string, port int) (string, error) {
 	app := commands.NewApp(build.Version, build.CommitSHA)
 
 	if arguments[0] == "scan" {
-		arguments = append(arguments, "--existing-worker=http://localhost:"+strconv.Itoa(port))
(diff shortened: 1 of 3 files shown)


cfabianski (commented):
Sounds like the `Quiet` options we wanted to introduce. What do you think @vjerci ?

vjerci:
> Sounds like the `Quiet` options we wanted to introduce. What do you think @vjerci ?

I agree with quiet option, It's less flags if we want to introduce more feedback logs.

cfabianski (approved):

cfabianski (commented):
Let's update https://app.delibr.com/app/7N672vhuYhBZiL5i6/documents/Lxrhbdjay9FvcBySE?filter=%7B%22bullet%22%3A%22m2tgho4pnnd66EMow%22%7D

PR #114, opened 2022-11-10
test: Add integration test for init command

spdawson:
## Description

Add an integration test covering the `init` command.

## Checklist

### DIFF
--- a/integration/flags/init_test.go
+++ b/integration/flags/init_test.go
@@ -7,11 +7,11 @@ import (
 )
 
 func TestInitCommand(t *testing.T) {
-	outputPath := "curio.yml"
-
-	testCase := testhelper.NewTestCase("init", []string{"init"})
-	testCase.OutputPath = outputPath
-	testCase.RunInTempDir = true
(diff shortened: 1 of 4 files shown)


integration/flags/init_test.go
```diff
@@ -0,0 +1,19 @@
...
+	"testing"
+
+	"github.com/bearer/curio/integration/internal/testhelper"
+)
+
+func TestInitCommand(t *testing.T) {
+	outputPath := "curio.yml"
+
+	testCase := testhelper.NewTestCase("init", []string{"init"})
+	testCase.OutputPath = outputPath
```
vjerci: think this properties shouldn't be public, but passed as params to NewTestCase.
I don't see us ever needing to mutate them or read them.
If NewTestCase func has too many params consider creating a struct for it
```go
testhelper.NewTestCase("init", []string{"init"},  testhelper.Test, testHelper.TestCaseRequest{
    OutputPath: outputPath,
    RuinInTempDir: true,
})
```

vjerci (approved):
good job

PR #119, opened 2022-11-14
test: Add integration tests covering scan format, report and output options

spdawson:
## Description

Adds snapshot-based integration tests for the following commands.

* `curio scan --format=json`
* `curio scan --format=yaml`
* `curio scan --report=detectors`
* `curio scan --report=dataflow`
* `curio scan --output ./some/path`
* `curio scan --context=health`
* `curio scan --disable-domain-resolution=true`
* `curio scan --skip-path="users/*.go,users/admin.sql"`

## Checklist


pkg/commands/scan.go
```diff
@@ -86,6 +86,10 @@ func NewScanCommand() *cobra.Command {
 }
 
 func readConfig(configFile string) error {
+	if configFile == "" {
```
spdawson: @vjerci This "fixes" the problem we saw with config file read errors when running all integration tests together. Not sure if this change is sensible; what do you think?

vjerci (dismissed):

pkg/commands/scan.go
vjerci: does this work:
https://github.com/spf13/viper/blob/v1.14.0/viper.go#L300
In theory calling it after every test might fix the bug

```
   viper.Reset()
```

spdawson: :bow: It does appear to work; thanks!

vjerci (approved):

PR #120, opened 2022-11-14
fix: update classification patterns

elsapet:
## Description
- Update classification JSON following updates in Rails app
- Also update scripts to include ID number in file name (this was the case for the old filenames, but the script did not reflect it)

## Checklist


elsapet:
Trying to keep the KPI PR small so I made these changes in a separate PR /cc @cfabianski

cfabianski (approved):

PR #122, opened 2022-11-15
chore: More minor output stream consolidation

spdawson:
## Description

* Use the Cobra standard error writer directly for the informational output of the `init` command
* Use the global error writer consistently for the progress bar: both the bar itself and the terminating newline should use the same writer

## Related

Follow-up for #105

## Checklist


cfabianski (commented):
Thank you for these little fixes!

cfabianski (approved):

PR #132, opened 2022-11-17
docs: Update CLI help text

markmichon:
## Description

Updates core help messaging and standardizes format for flag usage information.

## Todo
- [x] Update snapshots with new outputs

## Checklist

### DIFF
--- a/integration/flags/.snapshots/TestMetadataFlags-scan-help
+++ b/integration/flags/.snapshots/TestMetadataFlags-scan-help
@@ -12,13 +12,13 @@ Examples:
 
 
 Scan Flags
-      --context --context=health                                            Expand context of schema classification e.g. --context=health to include data types particular to health
-      --debug                                                               Enable debug logs
-      --disable-domain-resolution --disable-domain-resolution=true          Do not attempt to resolve detected domains during classification (default false), e.g. --disable-domain-resolution=true
-      --domain-resolution-timeout --domain-resolution-timeout=3s            Set timeout when attempting to resolve detected domains during classification (default 3 seconds), e.g. --domain-resolution-timeout=3s (default 3s)
-      --internal-domains --internal-domains="*.my-company.com,private.sh"   Define regular expressions for better classification of private or unreachable domains e.g. --internal-domains="*.my-company.com,private.sh"
-      --quiet                                                               Suppress non-essential messages
-      --skip-path --skip-path users/*.go,users/admin.sql                    Specify the comma separated files and directories to skip. Supports * syntax, e.g. --skip-path users/*.go,users/admin.sql
+      --context string                       Expand context of schema classification e.g., --context=health, to include data types particular to health
+      --debug                                Enable debug logs
+      --disable-domain-resolution            Do not attempt to resolve detected domains during classification (default false), e.g. --disable-domain-resolution=true
+      --domain-resolution-timeout duration   Set timeout when attempting to resolve detected domains during classification (default 3 seconds), e.g. --domain-resolution-timeout=3s (default 3s)
+      --internal-domains strings             Define regular expressions for better classification of private or unreachable domains e.g. --internal-domains="*.my-company.com,private.sh"
+      --quiet                                Suppress non-essential messages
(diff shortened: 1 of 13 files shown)


integration/flags/.snapshots/TestMetadataFlags-scan-help
```diff
@@ -1,24 +1,24 @@
...
-      --context string                       expand context of schema classification e.g. --context=health to include data types particular to health
-      --debug                                enable debug logs
-      --disable-domain-resolution            do not attempt to resolve detected domains during classification (default false), eg. --disable-domain-resolution=true
-      --domain-resolution-timeout duration   set timeout when attempting to resolve detected domains during classification (default 3 seconds), eg. --domain-resolution-timeout=TODO (default 3s)
-      --internal-domains strings             define regular expressions for better classification of private or unreachable domains eg. --internal-domains="*.my-company.com,private.sh"
-      --quiet                                suppress non-essential messages
-      --skip-path strings                    specify the comma separated files and directories to skip (supports * syntax), eg. --skip-path users/*.go,users/admin.sql
+      --context --context=health                                            Expand context of schema classification e.g. --context=health to include data types particular to health
```
cfabianski: something's fishy here

integration/flags/.snapshots/TestMetadataFlags-scan-help
```diff
@@ -1,24 +1,24 @@
...
+      --debug                                                               Enable debug logs
+      --disable-domain-resolution --disable-domain-resolution=true          Do not attempt to resolve detected domains during classification (default false), e.g. --disable-domain-resolution=true
+      --domain-resolution-timeout --domain-resolution-timeout=3s            Set timeout when attempting to resolve detected domains during classification (default 3 seconds), e.g. --domain-resolution-timeout=3s (default 3s)
+      --internal-domains --internal-domains="*.my-company.com,private.sh"   Define regular expressions for better classification of private or unreachable domains e.g. --internal-domains="*.my-company.com,private.sh"
+      --quiet                                                               Suppress non-essential messages
+      --skip-path --skip-path users/*.go,users/admin.sql                    Specify the comma separated files and directories to skip. Supports * syntax, e.g. --skip-path users/*.go,users/admin.sql
```
cfabianski: Something is fishy here. Is that really what you are getting when you run `curio scan help`?

pkg/commands/scan.go
```diff
@@ -39,13 +39,13 @@ var scanFlags = &flag.Flags{
...
 		Aliases: []string{"s"},
-		Short:   "Scan git repository",
-		Example: `  # Scan a local project including language-specific files
-  $ curio s /path/to/your_project
+		Short:   "Scan a directory or file",
+		Example: `  # Scan a local project, including language-specific files
+  $ curio scan /path/to/your_project
   # Scan a single file
-  $ curio s ./curio-ci-test/Pipfile.lock`,
+  $ curio scan ./curio-ci-test/Pipfile.lock`,
```
cfabianski: I don't think `Pipfile.lock` is a good example. Do you have something else in mind? Maybe a repo?
Also, we don't have `curio-ci-test` in the current directory so that makes it sort of useless. What do you think?

cfabianski (commented):
Left some comments but that's looking way better!

pkg/commands/scan.go
markmichon: What's a common use case for single-file scan? That would be the best option.

integration/flags/.snapshots/TestMetadataFlags-scan-help
markmichon: Looks like backticks aren't easily escapable in the flags, so I'm removing them. We'll live without code formating for now.

pkg/commands/scan.go
cfabianski: Not sure to understand what you mean here?

markmichon: Alright, I'm removing the single file example. If we find that it's confusing later, we can re-add with a more practical example.

cfabianski (approved):

PR #133, opened 2022-11-18
feat: Backport Rails framework classification

spdawson:
## Description

Create basic structure for framework classification. In this initial implementation, only the Rails framework is classified; the other frameworks will be classified in future work.

Sample YAML output for detectors report:

```yaml
- classification:
    Decision:
        reason: recipe_match
        state: valid
    recipe_match: true
    recipe_name: PostgreSQL
  detector_type: rails
  source:
    column_number: null
    filename: config/database.yml
    language: YAML
    language_type: data
    line_number: 85
    text: null
  type: framework_classified
  value:
    adapter: postgresql
    name: ""
```

## Checklist


pkg/classification/frameworks/frameworks.go
```diff
@@ -0,0 +1,99 @@
...
+
+type ClassifiedFramework struct {
+	*detections.Detection
+	Classification *Classification `json:"classification"`
+}
+
+type Classification struct {
+	RecipeMatch bool   `json:"recipe_match"`
+	RecipeName  string `json:"recipe_name,omitempty"`
+	Decision    classify.ClassificationDecision
```
cfabianski:
```suggestion
	Decision    classify.ClassificationDecision `json:"decision"`
```

cfabianski (dismissed):

elsapet (dismissed):

pkg/classification/dependencies/dependencies.go
```diff
@@ -44,7 +44,10 @@ func (classifier *Classifier) Classify(data detections.Detection) (*ClassifiedDe
 	var classification *Classification
 	value, ok := data.Value.(dependencies.Dependency)
 	if !ok {
-		return nil, errors.New("detection is not an dependency")
+		return &ClassifiedDependency{
+			Detection:      &data,
+			Classification: classification,
+		}, errors.New("detection is not an dependency")
```
cfabianski, line 50: Could you explain this change please?

pkg/classification/frameworks/frameworks.go
```diff
@@ -0,0 +1,99 @@
...
+
+type ClassifiedFramework struct {
+	*detections.Detection
+	Classification *Classification `json:"classification"`
+}
+
+type Classification struct {
+	RecipeMatch bool   `json:"recipe_match"`
+	RecipeName  string `json:"recipe_name,omitempty"`
+	Decision    classify.ClassificationDecision `json:"decision"`
```
cfabianski, line 20: You probably need to lint that line (sorry!)

cfabianski (commented):
Just one small question (@elsapet might be interested as well)
Otherwise, that looks good

pkg/classification/dependencies/dependencies.go
spdawson: We didn't previously have a test case that follows this code path; one is added in this PR. Following the pattern of the other "no match" cases, where we return a classified dependency, having a `nil` classification.

pkg/classification/frameworks/frameworks.go
spdawson: Linter seems to be passing... :confused:

pkg/classification/dependencies/dependencies.go
cfabianski: Cool 👍

cfabianski (dismissed):

pkg/report/writer/detectors.go
```diff
@@ -152,19 +154,20 @@ func (report *Detectors) AddFramework(
...
-		FrameworkType: frameworkType,
-		CommitSHA:     commitSHA,
-		Source:        source,
-		Value:         data,
-	})
+	if classifiedDetection.Source.LineNumber != nil {
+		classifiedDetection.CommitSHA = report.Blamer.SHAForLine(classifiedDetection.Source.Filename, *classifiedDetection.Source.LineNumber)
+	}
+
+	// @todo FIXME: Do we need to do anything with the `frameworkType`?
```
elsapet, line 168: In answer to the question here, I'm not sure - would we want to add it to the report? Maybe in the meantime, it's worth starting a small discussion for Delibr in terms of expected report output 🤔

elsapet (dismissed):
Looks good 👏

elsapet (dismissed):

pkg/report/writer/detectors.go
```diff
@@ -100,6 +100,8 @@ func (report *Detectors) AddSchema(
 	schema schema.Schema,
 	source source.Source,
 ) {
+	// @todo FIXME: Add classification here
```
elsapet, line 103: @cfabianski about this, I'm curious what the difference is between `AddSchema` and `AddDataType` (above)? Are we missing a classification step here?

cfabianski: Schema is for everything that is structured (SQL, GraphQL, OpenAPI, Protobuf)
DataType is for the detection in the code base `user.email`

cfabianski: We do miss a classification step here yeah

pkg/commands/process/worker/worker.go
```diff
@@ -29,7 +28,7 @@ func Start(port string) error {
 			var err error
 			json.NewDecoder(r.Body).Decode(&config) //nolint:all,errcheck
 
-			classifier, err = classsification.NewClassifier(&classsification.Config{Config: config})
+			classifier, err = classification.NewClassifier(&classification.Config{Config: config})
```
cfabianski, line 31: 😅

cfabianski (approved):

PR #141, opened 2022-11-18
chore: clean up and add data category UUID to data type

elsapet:
## Description
- Clean up `db.Datatype`
- Add data category UUID to data type, and use this to calculate policy severity instead of category name (more robust)

Changes: 

- output no longer includes category name or any UUIDs. 

Examples:

Policies
```
[
  {
    "result": [
      {
        "data_type": "Physical Address",
        "filename": "temp.rb",
        "line_number": 2,
        "policy_description": "Logger leaks detected",
        "policy_id": "detect_ruby_logger",
        "policy_name": "Logger leaks",
        "severity": "high"
      },
      {
        "data_type": "Unique Identifier",
        "filename": "temp.rb",
        "line_number": 2,
        "policy_description": "Logger leaks detected",
        "policy_id": "detect_ruby_logger",
        "policy_name": "Logger leaks",
        "severity": "critical"
      }
    ]
  }
]
```

Dataflow
```
{
  "data_types": [
    {
      "name": "Physical Address",
      "detectors": [
        {
          "name": "ruby",
          "locations": [
            {
              "filename": "temp.rb",
              "line_number": 2
            }
          ]
        }
      ]
    }
  ],
  "risks": [
    {
      "detector_id": "detect_ruby_logger",
      "data_types": [
        {
          "name": "Physical Address",
          "stored": false,
          "locations": [
            {
              "filename": "temp.rb",
              "line_number": 2
            }
          ]
        },
        {
          "name": "Unique Identifier",
          "stored": false,
          "locations": [
            {
              "filename": "temp.rb",
              "line_number": 2
            }
          ]
        }
      ]
    }
  ],
  "components": []
}
```

## Checklist


elsapet:
@cfabianski not sure if we want the category UUID in the reports - or, in the opposite direction, if we want both the category UUID and the category name to be included. Any guidance or should I ask in Delibr?

cfabianski:
> @cfabianski not sure if we want the category UUID in the reports - or, in the opposite direction, if we want both the category UUID and the category name to be included. Any guidance or should I ask in Delibr?

Let's leave a note in Delibr for sure but I think the report (in JSON) shouldn't include the UUID. This is more for internal purposes. I think we simply need to do something like `json:"-"` will do the trick so that we can use the UUID to have a more robust reconciliation but we would not display them in the report.
Does that make sense?

cfabianski (dismissed):
Apart from my comment, that looks good to me

cfabianski:
Q. Why do we have 2 datatypes for this line?
Can you maybe share the content of the `temp.rb` to double check that?

elsapet:
@cfabianski temp.rb looks like this

```
logger.info(user.address)
```

cfabianski:
> @cfabianski temp.rb looks like this
> 
> ```
> logger.info(user.address)
> ```

So in that case, I don't really understand this 

```
"data_type": "Unique Identifier",
```

😕

cfabianski (approved):

PR #148, opened 2022-11-21
feat: Add UUIDs to recipes and include these in the report

spdawson:
## Description

Relying on the recipe name is fragile; switch to use the recipe UUID, and output this in reports for internal consumption.

#### Before

```yaml
- classification:
    decision:
        reason: recipe_match
        state: valid
    recipe_match: true
    recipe_name: PostgreSQL
  detector_type: rails
  source:
    column_number: null
    filename: config/database.yml
    language: YAML
    language_type: data
    line_number: 85
    text: null
  type: framework_classified
  value:
    adapter: postgresql
    name: ""
```

#### After

The `recipe_uuid` attribute is present:

```yaml
- classification:
    decision:
        reason: recipe_match
        state: valid
    recipe_match: true
    recipe_name: PostgreSQL
    recipe_uuid: 428ff7dd-22ea-4e80-8755-84c70cf460db
  detector_type: rails
  source:
    column_number: null
    filename: config/database.yml
    language: YAML
    language_type: data
    line_number: 85
    text: null
  type: framework_classified
  value:
    adapter: postgresql
    name: ""
```

#### Before

```yaml
datatypes:
    - name: Email Address
      detectors:
        - name: ruby
          locations:
            - filename: config/initializers/devise.rb
              linenumber: 186
risks: []
components:
    - name: aws key management service (kms)
      locations:
        - detector: gemfile-lock
          filename: Gemfile.lock
          linenumber: 78
    - name: aws s3
      locations:
        - detector: gemfile-lock
          filename: Gemfile.lock
          linenumber: 81
    - name: postgresql
      locations:
        - detector: gemfile-lock
          filename: Gemfile.lock
          linenumber: 151
    - name: redis
      locations:
        - detector: gemfile-lock
          filename: Gemfile.lock
          linenumber: 191
```

#### After

Same as before (no change).

## Checklist

### DIFF
--- a/pkg/report/output/dataflow/components/components.go
+++ b/pkg/report/output/dataflow/components/components.go
@@ -12,10 +12,12 @@ import (
 
 type Holder struct {
 	components map[string]*component // group components by name
+	isInternal bool
 }
 
 type component struct {
 	name      string
+	uuid      string
 	detectors map[string]*detector // group detectors by detectorName
 }
 type detector struct {
@@ -27,9 +29,10 @@ type fileHolder struct {
 	lineNumbers map[int]int //group lines by linenumber
 }
 
-func New() *Holder {
+func New(isInternal bool) *Holder {
 	return &Holder{
 		components: make(map[string]*component),
+		isInternal: isInternal,
 	}
 }
 
@@ -44,7 +47,7 @@ func (holder *Holder) AddInterface(detection interface{}) error {
 	}
 
 	if value.Classification.Decision.State == classify.Valid {
-		holder.addComponent(strings.ToLower(value.Classification.RecipeName), string(value.DetectorType), value.Source.Filename, *value.Source.LineNumber)
+		holder.addComponent(strings.ToLower(value.Classification.RecipeName), value.Classification.RecipeUUID, string(value.DetectorType), value.Source.Filename, *value.Source.LineNumber)
 	}
 
 	return nil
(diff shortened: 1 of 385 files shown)


pkg/report/output/dataflow/types/components.go
```diff
@@ -1,7 +1,7 @@
 package types
 
 type Component struct {
-	Name      string
+	Uuid      string
```
elsapet: Currently in the codebase, we either have `UUID` or `uuid` (for private). Suggestion to prefer `UUID` for consistency here. 

```suggestion
	UUID      string
```

pkg/report/output/dataflow/components/components.go
```diff
@@ -61,23 +61,23 @@ func (holder *Holder) AddDependency(detection interface{}) error {
...
-		holder.addComponent(strings.ToLower(value.Classification.RecipeName), string(value.DetectorType), value.Source.Filename, *value.Source.LineNumber)
+		holder.addComponent(strings.ToLower(value.Classification.RecipeUUID), string(value.DetectorType), value.Source.Filename, *value.Source.LineNumber)
 	}
 
 	return nil
 }
 
 // addComponent adds component to hash list and at the same time blocks duplicates
-func (holder *Holder) addComponent(componentName string, detectorName string, fileName string, lineNumber int) {
+func (holder *Holder) addComponent(componentUuid string, detectorName string, fileName string, lineNumber int) {
```
elsapet:
```suggestion
func (holder *Holder) addComponent(componentUUID string, detectorName string, fileName string, lineNumber int) {
```

elsapet (dismissed):
Looks good to me, but I wonder if we want to the recipe name as well as recipe UUID in the report. @cfabianski any thoughts on this?

cfabianski:
> Looks good to me, but I wonder if we want to the recipe name as well as recipe UUID in the report. @cfabianski any thoughts on this?

You are on the right track @elsapet 
There is a difference between internal and external. @elsapet you have done something similar, wanna show @spdawson?

cfabianski (approved):
LGTM

PR #157, opened 2022-11-23
feat: application level encryption

vjerci:
## Description
This Pr adds verified by, It adds risk type and adds support for rego policies as processor step between detectors and dataflow.
To dataflow report it adds non mandatory field encrypted, and non mandatory field verified by.

### example output

```
data_types:
    - name: Country
      detectors:
        - name: detect_sql_create_public_table
          locations:
            - filename: schema.sql
              line_number: 8
              encrypted: true
              verified_by:
                - detector: detect_encrypted_ruby_class_properties
                  filename: user.rb
                  line_number: 2
        - name: ruby
          locations:
            - filename: user.rb
              line_number: 2
    - name: Date of birth
      detectors:
        - name: detect_sql_create_public_table
          locations:
            - filename: schema.sql
              line_number: 6
              encrypted: true
              verified_by:
                - detector: tanker_encrypted
    - name: Email Address
      detectors:
        - name: detect_sql_create_public_table
          locations:
            - filename: schema.sql
              line_number: 5
              encrypted: true
              verified_by:
                - detector: detect_encrypted_ruby_class_properties
                  filename: user.rb
                  line_number: 2
        - name: ruby
          locations:
            - filename: user.rb
              line_number: 2
    - name: Firstname
      detectors:
        - name: detect_sql_create_public_table
          locations:
            - filename: schema.sql
              line_number: 3
    - name: Lastname
      detectors:
        - name: detect_sql_create_public_table
          locations:
            - filename: schema.sql
              line_number: 4
    - name: Physical Address
      detectors:
        - name: detect_sql_create_public_table
          locations:
            - filename: schema.sql
              line_number: 7
              encrypted: true
              verified_by:
                - detector: detect_encrypted_ruby_class_properties
                  filename: user.rb
                  line_number: 2
        - name: ruby
          locations:
            - filename: user.rb
              line_number: 2
components: []

```

## Related
https://bearer.atlassian.net/browse/AMA-3230
https://bearer.atlassian.net/browse/AMA-3231
https://bearer.atlassian.net/browse/AMA-3232
https://bearer.atlassian.net/browse/AMA-3233
https://bearer.atlassian.net/browse/AMA-3234

## Checklist


integration/flags/.snapshots/TestInitCommand-init
```diff
@@ -8,14 +8,112 @@ report:
...
+                  CREATE TABLE public.$TABLE_NAME (
+                    <$COLUMN>
+                  )
+            param_parenting: true
+            processors:
+                - query: |
+                    verified_by = data.bearer.encrypted_tanker.verifiedBy
+                    encrypted = data.bearer.encrypted_tanker.encrypted
+                  modules:
+                    - name: bearer.encrypted_tanker
```
cfabianski: can we use something less "static"? I thought we would be using a more generic module for it and pass the prefix as an argument?

integration/flags/.snapshots/TestInitCommand-init
```diff
@@ -8,14 +8,112 @@ report:
...
+                      content: |-
+                        package bearer.encrypted_tanker
+
+                        import future.keywords
+
+
+                        default encrypted := false
+
+                        encrypted = true {
+                            startswith(lower(input.target.value.field_name),  "tanker_encrypted")
```
cfabianski: Let's not have this one as part of the default (let's just try with this)

integration/flags/.snapshots/TestInitCommand-init
```diff
@@ -8,14 +8,112 @@ report:
...
+                      content: |-
+                        package bearer.encrypted_verified
+
+                        import future.keywords
+
+
+                        default encrypted := false
+
+
+                        rubyEncrypted[location] {
```
cfabianski: I don't know what we should be using here camelCase or snake_case
It feels like it should be snake_case. Everything else seems to be snake_case otherwise. WDYT?

cfabianski: cc @elsapet

integration/flags/testdata/verified_by/schema.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE public.Users (
+  id bigint NOT NULL,
+  first_name character varying NOT NULL,
+  last_name character varying NOT NULL,
+  email character varying NOT NULL,
+  tanker_encrypted_date_of_birth character varying NOT NULL,
+  city character varying NOT NULL,
+  country character varying NOT NULL,
+);
```
cfabianski, line 9: Do you know if this is going to work out of the box or if we need to use the same workaround we have in `sql detector` where we split every time we have a `CREATE TABLE`?

cfabianski (commented):
Looking great already! Some comments

integration/flags/.snapshots/TestInitCommand-init
vjerci: Could you clarify it, not sure i understand the question.
Is it that you would like to use `data.encrypted_tanker` instead of `data.bearer.encrypted_tanker`?
It is rego requirement to put some kind of name there and use it in query.

integration/flags/.snapshots/TestInitCommand-init
vjerci: think this comment is missing example.

integration/flags/.snapshots/TestInitCommand-init
vjerci: I've chosen cameCase for variable names in order to make distinction between variable and json fields.
opa does use snake case for everything in examples

integration/flags/testdata/verified_by/schema.sql
vjerci: It doesn't work out of the box. At the moment we don't have that workaround implemented for custom detectors. However if we implement it for custom detectors it's gonna work.

integration/flags/.snapshots/TestInitCommand-init
cfabianski: Could we use something like a generic (built-in) module (e.g. matcher?) that would accept an argument that we would use to match against.

integration/flags/testdata/verified_by/schema.sql
cfabianski: ok Let's make sure we have a card for it :)

integration/flags/.snapshots/TestInitCommand-init
vjerci: We could pass in "tanker_encrypted" in input of a lets call it prefix checker module, then have a module which checks if field has a prefix based on that prametrised input.
But honestly to me that feels like a wrong thing to do, because I don't think we can forsee most of the possible use cases, some users will have it as prefix some as multiple different suffixes. Some will use completely different logic.
To me it feels better to nudge users to build their own processors (or we build them for them) instead of passing params to pre built ones.

integration/flags/testdata/verified_by/schema.sql
vjerci: https://app.delibr.com/app/7N672vhuYhBZiL5i6/documents/Lxrhbdjay9FvcBySE?zoom=4sgZ4mmKww6McBD5A

integration/flags/.snapshots/TestInitCommand-init
vjerci: remove tanker encrypted and save it in notion.

integration/flags/.snapshots/TestInitCommand-init
vjerci: do snake case

integration/flags/.snapshots/TestInitCommand-init
```diff
@@ -8,14 +8,90 @@ report:
...
+            languages:
+                - ruby
+            patterns:
+                - |
+                  class $CLASS_NAME < ApplicationRecord
+                    encrypts <$ARGUMENT>
+                  end
+            param_parenting: true
+            processors: []
+            singularilize: false
```
cfabianski:
```suggestion
            singularize: false
```

integration/flags/.snapshots/TestInitCommand-init
```diff
@@ -8,14 +8,90 @@ report:
...
+            lowercase: false
```
cfabianski: shouldn't this be?
```suggestion
            lowercase: true
```

pkg/commands/process/settings/settings.go
```diff
@@ -26,28 +27,50 @@ var LevelHigh = "high"
...
 	Disabled       bool
+	Type           string
 	Languages      []string
 	Patterns       []string
 	ParamParenting bool `yaml:"param_parenting"`
-	Metavars       map[string]MetaVar
-	Stored         bool
+	Processors     []Processor
+
+	Singularilize bool
```
cfabianski:
```suggestion
	Singularize bool
```

pkg/detectors/custom/config/config.go
```diff
@@ -8,6 +8,8 @@ type CompiledRule struct {
 	Params         []Param
 	Metavars       map[string]settings.MetaVar
 	ParamParenting bool
+	Singularize    bool
```
cfabianski: It's correctly spelled here 🤔

pkg/detectors/custom/custom.go
```diff
@@ -71,6 +74,8 @@ func (detector *Detector) CompileRules(rulesConfig map[string]settings.Rule) err
 				compiledRule.RuleName = ruleName
 				compiledRule.Metavars = rule.Metavars
 				compiledRule.ParamParenting = rule.ParamParenting
+				compiledRule.Lowercase = rule.Lowercase
+				compiledRule.Singularize = rule.Singularilize
```
cfabianski: Ah ok! 🚨

pkg/report/output/policies/policies.go
```diff
@@ -32,42 +32,31 @@ type PolicyResult struct {
...
-					Dataflow:       dataflow,
-					DataCategories: db.Default().DataCategories,
-				},
-			),
-		)
 		if err != nil {
 			return nil, err
 		}
 
 		log.Debug().Msgf("result %#v", rs)
```
cfabianski: Let's get rid of that

cfabianski (commented):
I don't see the policy using this info. Is that coming in a separate PR?

vjerci:
> I don't see the policy using this info. Is that coming in a separate PR?

yeah something is wrong my branches got mixed. fixing it.

integration/flags/.snapshots/TestInitCommand-init
vjerci: resolving it as it was fixed on separate branch, i've added commits to it.

pkg/detectors/custom/config/config.go
vjerci: resolving it as it was fixed on separate branch, i've added commits from it.

pkg/report/output/policies/policies.go
vjerci: done

cfabianski (approved):

PR #174, opened 2022-11-28
feat: add support for variable reconciliation

vjerci:
## Description
This pr adds variable reconciliation, as well as refractors ruby datatypes to work with node instead of with tree.
Detections got from variable reconciliations are duplicated for each custom detection occurence

## Checklist


vjerci (commented):
I still need to make it a bit more efficient, because at the moment for each custom detection occurence all datatypes get pulled. I want to do it once per file. It is low hanging fruit

cfabianski:
Couple of examples where I would have expected it to get me results but doesn't

```ruby
payload = {
	user: {
		first_name: "John",
		last_name: "Doe",
	}
}

token = JWT.encode payload, nil, 'none'
```

```ruby
User = Struct.new(:first_name, :last_name, keyword_init: true)
user = User.new(first_name: "first", last_name: "last")

response = Curl.post("http://my.api.com/users/create", user)
```

cfabianski (approved):

PR #175, opened 2022-11-29
fix: Use full file paths in dataflow report

spdawson:
## Description

All filenames in the output should consistently use the full path, including the scan target directory prefix.

### Sample output before

```yaml
data_types:
    - name: Email Address
      detectors:
        - name: rails
          locations:
            - filename: ../scan_target_rails/db/schema.rb
              line_number: 18
        - name: ruby
          locations:
            - filename: ../scan_target_rails/config/initializers/devise.rb
              line_number: 186
components:
    - name: Disk
      locations:
        - detector: rails
          filename: config/storage.yml
          line_number: 5
    - name: PostgreSQL
      locations:
        - detector: rails
          filename: config/database.yml
          line_number: 85
    - name: Redis
      locations:
        - detector: gemfile-lock
          filename: Gemfile.lock
          line_number: 191
```

### Sample output after

```yaml
data_types:
    - name: Email Address
      detectors:
        - name: rails
          locations:
            - filename: ../scan_target_rails/db/schema.rb
              line_number: 18
        - name: ruby
          locations:
            - filename: ../scan_target_rails/config/initializers/devise.rb
              line_number: 186
components:
    - name: Disk
      locations:
        - detector: rails
          filename: ../scan_target_rails/config/storage.yml
          line_number: 5
    - name: PostgreSQL
      locations:
        - detector: rails
          filename: ../scan_target_rails/config/database.yml
          line_number: 85
    - name: Redis
      locations:
        - detector: gemfile-lock
          filename: ../scan_target_rails/Gemfile.lock
          line_number: 191
```

## Checklist

### DIFF
--- a/pkg/report/output/dataflow/dataflow.go
+++ b/pkg/report/output/dataflow/dataflow.go
@@ -11,6 +11,7 @@ import (
 	"github.com/bearer/curio/pkg/report/detections"
 	"github.com/bearer/curio/pkg/report/output/dataflow/components"
 	"github.com/bearer/curio/pkg/report/output/dataflow/datatypes"
+	"github.com/bearer/curio/pkg/report/output/dataflow/detectiondecoder"
 	"github.com/bearer/curio/pkg/report/output/dataflow/risks"
 
 	"github.com/bearer/curio/pkg/report/output/dataflow/types"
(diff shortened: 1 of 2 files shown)


cfabianski (commented):
what do you think @elsapet?

pkg/report/output/dataflow/dataflow.go
```diff
@@ -110,17 +112,17 @@ func GetOutput(input []interface{}, config settings.Config, isInternal bool) (*D
...
 
 		case detections.TypeDependencyClassified:
-			err := componentsHolder.AddDependency(detection)
+			err := componentsHolder.AddDependency(detection, fullFilename)
 			if err != nil {
 				return nil, err
 			}
 		case detections.TypeInterfaceClassified:
-			err := componentsHolder.AddInterface(detection)
+			err := componentsHolder.AddInterface(detection, fullFilename)
```
elsapet: It feels confusing to pass a filename to the `AddX` methods in addition, because the detection itself also has a filename. I'd prefer to cast the detection first (i.e. in the `AddInterface` example. move `value, err := detectiondecoder.GetClassifiedInterface(detection)` out of the `AddInterface`). update its Source.Filename, and pass the updated detection to the `AddX` method.

In this way, `AddX` continues to take just a detection, which is more straightforward to me.

Any thoughts?

elsapet (approved):
🎉

pkg/report/output/dataflow/dataflow.go
spdawson: Thanks @elsapet :bowing_man: I was trying to keep the change as small as possible, originally; but I think it's better with your suggestion incorporated :slightly_smiling_face:

PR #184, opened 2022-12-01
fix: use parent from the location not from the datatype

cfabianski:
## Description

We were using only 1 context per custom detection / datatype while the context should be at the location level

Also removing the datatype detection in the risk as they would basically be duplicated 

## Checklist


elsapet (approved):
🙏

didroe (approved):

elsapet:
@spdawson FYI once this PR is merged, we should be able to get the code extract (i.e. parent context) for the SMTP risks :)

spdawson (approved):

PR #185, opened 2022-12-01
feat: some config fields are not loading

vjerci:
## Description
This pr fixes bug where we weren't using config from file when file was specified.
Viper is actually using [mapstructure lib](https://github.com/mitchellh/mapstructure) behind the scenes, which means we need to add mapstructure metatags for settings being loaded from file

## Checklist


pkg/commands/process/settings/settings.go
```diff
@@ -16,7 +16,7 @@ type Config struct {
 	Worker         flag.WorkerOptions `json:"worker" yaml:"worker"`
 	Scan           flag.ScanOptions   `json:"scan" yaml:"scan"`
 	Report         flag.ReportOptions `json:"report" yaml:"report"`
-	CustomDetector map[string]Rule    `json:"custom_detector" yaml:"custom_detector"`
+	CustomDetector map[string]Rule    `mapstructure:"custom_detector" json:"custom_detector" yaml:"custom_detector"`
```
cfabianski, line 19: I don't understand why this is required only for custom_detector. Could you explain a bit the reasoning behind please?

vjerci: It's required only for 2 letter words, but honestly i wanted to add it to all fields just seems that i've skipped some of them let me recheck and fix it.

vjerci: done

cfabianski: https://github.com/Bearer/curio/pull/185/commits/3cecc1574be78df828fb30c7b644ea28fad84b6d that is the actual fix though right?

vjerci: Previously default custom detectors were always loaded no matter what you specified in config file.
this hash fixes that default custom detectors are not always loaded, you only have custom detectors specified in the file if you use `--config-file`.
`mapstructure:"..."` fixes that custom detector properties are correctly loaded, previously they were loaded expect for multi word properties which were just default values.

cfabianski (approved):

PR #187, opened 2022-12-01
feat: Ruby policy for insecure communication (SSL not enforced)

spdawson:
## Description

Add custom detection and associated policy, for failure to enforce SSL in a Rails application, when the application processes sensitive data.

Also: reworks the "insecure SMTP" Rails policy, to make this conditional on the detection of sensitive data.

## Checklist


pkg/commands/process/settings/policies.yml
```diff
@@ -47,3 +47,12 @@ insecure_smtp_processing_sensitive_data:
   modules:
     - path: policies/insecure_smtp.rego
       name: bearer.insecure_smtp
+insecure_communication_processing_sensitive_data:
+  description: "Insecure communication in an application processing sensitive data"
```
elsapet, line 70: Do we also need to confirm that sensitive data types have been detected in the scan? 
Wondering because this is quite similar to my current policy 🤔  https://bearer.atlassian.net/browse/AMA-3287

spdawson: :+1: Ah, yes; you're right.

cfabianski (approved):

PR #205, opened 2022-12-06
feat: narrow down the scope of variable reconciliation

vjerci:
## Description
This pr narrows down the scope of variable reconciliation, as described in test case and discussed on call.

## Checklist


pkg/parser/datatype/merge.go
```diff
@@ -31,3 +32,34 @@ func MergeDatatypesByPropertyNames(source datatype.DataTypable, enrichment datat
 		sourceProperties[nodeID] = enrichmentDatatype
 	}
 }
+
+// CloneDeepestProperties deep clones source and iterates trough various leveles of same name source and enrichement properties adding enrichment properties to the edge of source only when source edge doesn't have any properties
```
didroe: You could wrap this to make it a bit easier to read:

```suggestion
// CloneDeepestProperties deep clones source and iterates trough various levels 
// of same name source and enrichement properties adding enrichment properties 
// to the edge of source only when source edge doesn't have any properties
```

didroe (approved):

pkg/parser/datatype/merge.go
vjerci: wasn't sure if golang docs picks it up if its a multiline.

PR #217, opened 2022-12-08
feat: fix custom detector bug related to data extraction from session policy

vjerci:
## Description
This pr adds support for generic data extraction used in session and cookies policies, it might help with some other policies too since most of the work is done on adding support for various arguments extractions.
It also refractors custom detector tests so that they are more streamlined.

for what types of data is supported check out `./pkg/detectors/custom/testdata/ruby/sessions`

## Checklist

- [] I've included usage information in the description if CLI behavior was updated or added.


pkg/detectors/ruby/custom_detector/extract_arguments.go
```diff
@@ -44,36 +45,102 @@ func (detector *Detector) extractArguments(node *parser.Node, idGenerator nodeid
...
+		if parserdatatype.IsParentedByNodeID(node.ID(), target.Node) {
+			joinedDatatypes[nodeID] = target
+		}
 	}
 
 	return joinedDatatypes, nil
+
+}
+
+func extractHash(datatypes map[parser.NodeID]*schemadatatype.DataType, node *parser.Node) {
```
vjerci, line 109: this will extract object with empty object name and filled in field names of 1 level depth (as specified in test case `hash`)

pkg/detectors/ruby/datatype/properties.go
```diff
@@ -134,11 +135,8 @@ func addProperties(node *parser.Node, helperDatatypes map[parser.NodeID]*schemad
 					}
 
 					key := pair.ChildByFieldName("key")
-					if key.Type() != "hash_key_symbol" {
```
vjerci, line 137: I've removed this so that both cases are supported trough variable reconciliation, According to tests it doesn't break anything.
```
user = {"first_name": "Mark"}
```
and case
```
user = {first_name: "Mark"}
```

pkg/detectors/ruby/custom_detector/extract_arguments.go
```diff
@@ -44,36 +45,102 @@ func (detector *Detector) extractArguments(node *parser.Node, idGenerator nodeid
...
-			continue
+
+			joinedDatatypes[nodeID] = target
 		}
+
+		return joinedDatatypes, nil
 	}
 
-	complexDatatypes := datatype.Discover(node, idGenerator)
+	complexDatatypes := datatype.Discover(node.Parent(), idGenerator)
```
vjerci, line 93: If we don't select parent here datatype extraction returns nothing for calls.
```
  session[:current_user] = user.email
```

PR #220, opened 2022-12-09
fix: performance on large repositories

cfabianski:
## Description

Fixes performance issues seen on the Gitlab repository. 

The cause was running the encrypted field processor for each detection. We now run the processor once with all detections.

Also:
- Removes configurable processors, in practice the code only supported the encrypted field one anyway.
- Adds some additional logging
- Ignores symlinks
- Increases max file size to 2mb

AMA-3332

## Checklist


didroe:
https://bearer.atlassian.net/browse/AMA-3332

PR #229, opened 2022-12-12
ci: fix doc generation

gotbadger:
## Description
Command was removed and doc page was not updated we should try and see if we can automate this in the future to avoid such bugs.

## Checklist


markmichon:
@gotbadger Yeah unfortunately that array in the template has to exist to control the order of command output, unless there's a way to tell the go script to prepend them with numbers or some other clear indicator.

PR #237, opened 2022-12-13
fix: Include correct indentation in first line of policy breach context

spdawson:
## Description

Previously, we would omit any and all content that precedes the match, on the first line of the match context. In particular, if the first line of a match has initial whitespace, then this would be trimmed in the output.

Unfortunately, in order to reconstruct the first line fully, we must re-read it from the originating file. This is done using a
line scanner, to avoid reading the entire file into memory. Nevertheless, this is sub-optimal; in the future, it may be worthwhile improving on this implementation.

Also: show only a single-line code extract, when the policy dictates that parent context be omitted.

#### Before

Before

#### After

After

## Checklist

### DIFF
--- a/integration/flags/.snapshots/TestInitCommand-init
+++ b/integration/flags/.snapshots/TestInitCommand-init
@@ -82,7 +82,7 @@ scan:
             metavars: {}
             stored: false
             detect_presence: true
-            omit_parent: true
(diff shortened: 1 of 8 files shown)


elsapet (approved):
🙌

PR #239, opened 2022-12-13
chore: Rename policy breach -> policy failure

spdawson:
## Description

Rename "policy breach" to "policy failure" throughout the codebase.

## Checklist

### DIFF
--- a/pkg/report/output/policies/policies.go
+++ b/pkg/report/output/policies/policies.go
@@ -89,7 +89,7 @@ func GetOutput(dataflow *dataflow.DataFlow, config settings.Config) (map[string]
 				return nil, err
 			}
 
-			for _, policyOutput := range policyResults["policy_breach"] {
+			for _, policyOutput := range policyResults["policy_failure"] {
 				policyResult := PolicyResult{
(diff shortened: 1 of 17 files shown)


pkg/report/output/policies/policies.go
```diff
@@ -124,7 +124,7 @@ func BuildReportString(policyResults map[string][]PolicyResult, policies map[str
 
 	writePolicyListToString(reportStr, policies)
 
-	breachedPolicies := map[string]map[string]bool{
+	failedPolicies := map[string]map[string]bool{
```
elsapet: Wondering if `policyFailures` makes more sense and aligns better with `policy_failure` that we use elsewhere. What do you think? 

I probably should have used `policyBreaches` here for the same reason... 😅 

```suggestion
	policyFailures := map[string]map[string]bool{
```

elsapet (approved):
Small suggestion you can take or leave :) 
Looks great!! 👏

pkg/report/output/policies/policies.go
spdawson: :+1: Good idea

PR #242, opened 2022-12-13
fix: dont show config load when no config given

gotbadger:
## Description

Currently you get `Loaded  configuration file` when no file is given this PR hides the message when no config is set

## Checklist


cfabianski (approved):
Thank you!

PR #244, opened 2022-12-13
feat: password length too weak in an application processing data category

vjerci:
## Description
This pr adds support for password length to weak policy.
It adds support in custom detector for different types of filters (`minimum` `maximum`) that match when there is violation as opposed to values matching which match when there is occurence of string.
other major difference in behaviour is that with values we probabbly want to match  cases of all occurences of variables with values while with min and max we want to match if there was a single breach.

## Checklist


integration/custom_detectors/custom_detectors_test.go
```diff
@@ -25,6 +25,7 @@ func TestCustomDetectors(t *testing.T) {
 		newScanTest("ruby", "ssl_certificate_verification_disabled", "ssl_certificate_verification_disabled.rb"),
 		newScanTest("ruby", "ruby_http_detection", "ruby_http_detection.rb"),
 		newScanTest("ruby", "ruby_weak_password_encryption", "weak_password_encryption.rb"),
+		newScanTest("ruby", "detect_rails_jwt", "detect_password_length.rb"),
```
cfabianski:
```suggestion
		newScanTest("ruby", "detect_password_length", "detect_password_length.rb"),
```

pkg/commands/process/settings/settings.go
```diff
@@ -56,13 +56,16 @@ func (modules Modules) ToRegoModules() (output []rego.Module) {
 }
 
 type PatternFilter struct {
-	Variable string
-	Values   []string
+	Variable       string   `mapstructure:"variable" json:"variable" yaml:"variable"`
+	Values         []string `mapstructure:"values" json:"values" yaml:"values"`
+	Minimum        *int     `mapstructure:"minimum" json:"minimum" yaml:"minimum"`
+	Maximum        *int     `mapstructure:"Maximum" json:"Maximum" yaml:"Maximum"`
```
cfabianski:
```suggestion
	Maximum        *int     `mapstructure:"maximum" json:"maximum" yaml:"maximum"`
```

cfabianski (approved):

PR #257, opened 2022-12-15
fix: add test for policy flags

elsapet:
## Description
Adds an integration test for policy flags

## Checklist


cfabianski (approved):
💌

PR #258, opened 2022-12-15
feat: Consolation message for codebases where no policies supported

spdawson:
## Description

When a codebase is such &mdash; by virtue of languages, technologies, etc. present &mdash; that no policies are applicable, then we should show a message summarizing what we detected.

### Limitations

In this initial implementation, the following features are missing.

* No warning icon/emoji displayed
* No count of database, and associated data type counts, displayed
* No count of external APIs displayed
* Internal API counting is implemented, but only in a rudimentary manner: components whose name string begins with either `"http://"` or `"https://"` are counted as internal APIs

### Screenshot

Screenshot_2022-12-15_10-21-01

## Checklist

### DIFF
--- a/pkg/report/output/output.go
+++ b/pkg/report/output/output.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/bearer/curio/pkg/commands/process/settings"
 	"github.com/bearer/curio/pkg/flag"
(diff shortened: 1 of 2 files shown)


pkg/report/output/output.go
```diff
@@ -16,23 +17,52 @@ import (
...
+	if err != nil {
+		return
+	}
+
+	if !reportSupported {
+		placeholderStr, err2 := getPlaceholderOutput(report, config, lineOfCodeOutput)
+		if err2 != nil {
+			err = err2
+			return
+		}
```
elsapet: Minor, but would something like this work? 

```suggestion
		var placeholderStr *strings.Builder
		placeholderStr, err = getPlaceholderOutput(report, config, lineOfCodeOutput)
		if err != nil {
			return
		}
```

pkg/report/output/stats/stats.go
```diff
@@ -1,20 +1,41 @@
 package stats
 
 import (
+	"fmt"
+	"strings"
+
+	"github.com/bearer/curio/pkg/classification/db"
 	"github.com/bearer/curio/pkg/commands/process/settings"
+
+	// reportdetectors "github.com/bearer/curio/pkg/report/detectors"
```
elsapet: Let's remove this for now, to avoid confusion. The compiler won't let us forget it when we return to the uncommented code 👮 

```suggestion
```

elsapet (approved):
Two small suggestions, no blockers ✅

pkg/report/output/output.go
spdawson: That does work, thanks!

PR #271, opened 2022-12-16
fix: Temp file path on Linux is missing separator

spdawson:
## Description

The malformed temporary file path causes the scan to fail. Sample error message:

```
filesystem scan error: open /tmpe566c0f8c7ffb00cb50b8bc7fa842e58801f054d-devSHA.jsonl: permission denied
```

## Checklist


cfabianski (approved):

cfabianski:
Could be what was making the [CI hang](https://github.com/Bearer/curio/pull/271)

The error might have been swallowed and the process wasn't exiting properly. @vjerci

Issue #266, opened 2022-12-15
CI hangs when running the Unit Tests


cfabianski:
## Description

The CI seems to hang when running the integration tests.
This PR https://github.com/Bearer/curio/pull/250 attempted to address this but this behaviour seems to still be happening 

See this actions https://github.com/Bearer/curio/actions/runs/3708056420/jobs/6285192136

## Expected Behavior

The CI is reliable

## Actual Behavior

The CI hangs

## Possible Fix

- [x] Separate integration tests and Unit Tests from the CI
- [ ] Fix the integration tests on GitHub Action 

## Steps to Reproduce

1. Issue a PR
2. Check the CI 

## Context

Only seems to happen on GitHub Actions.
I don't seem to be able to reproduce it locally for now.

cfabianski:
https://github.com/Bearer/curio/pull/271#issuecomment-1354530395

cfabianski:
The CI doesn't hang anymore. Closing

PR #276, opened 2022-12-16
feat: harden integration tests

vjerci:
## Description
This pr sets a default timeout of 1min per integration tests.
If the timeout is reached without command finishing its execution the test case should fail.

The goal is to have certain test fail and prevent hanging the whole test suite.

## Related
https://github.com/Bearer/curio/issues/266

## Checklist

### DIFF
--- a/integration/flags/init_test.go
+++ b/integration/flags/init_test.go
@@ -12,7 +12,7 @@ func TestInitCommand(t *testing.T) {
 	filePath := testhelper.GetCWD() + "/curio.yml"
 	arguments := []string{"init"}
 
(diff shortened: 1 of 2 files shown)


spdawson (approved):
:muscle:

cfabianski:
That's cool! Do you know if this is going to be enough to return the error if an issue is happening on the worker side?

vjerci:
> That's cool! Do you know if this is going to be enough to return the error if an issue is happening on the worker side?

@cfabianski I've tested it by adding sleep(10*time.Minute) to worker, and it successfully killed the test.

Issue #296, opened 2022-12-19
Curio doesn't run on vanilla CloudLinux, Centos & Debian vanilla setup with install script. (missing libs)


b1pb1p:
## Description

Installation script isn't sufficient to have curio working on some Linux distributions 

## Actual Behavior

```
[centos@centos-brr-test ~]$ ./bin/curio scan bear-publishing/
./bin/curio: /lib64/libstdc++.so.6: version `CXXABI_1.3.9' not found (required by ./bin/curio)
./bin/curio: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by ./bin/curio)
./bin/curio: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by ./bin/curio)
./bin/curio: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by ./bin/curio)
```

```
[cloudlinux@archlinux-brr-test ~]$ ./bin/curio scan bear-publishing/
./bin/curio: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by ./bin/curio)
./bin/curio: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by ./bin/curio)
```

```
debian@debion-brr-test:~$ ./bin/curio scan bear-publishing/
./bin/curio: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ./bin/curio)
./bin/curio: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./bin/curio)
```

## Possible Fix

Make the install script install the required libs ? 

## Steps to Reproduce

Try installing curio on Debian / ArchiLniux / CentOS vanilla. 

## Context

QA Test

## Your Environment

- Version used: v0.19
- Target codebase stack (e.g. Ruby): bear-publishing
- Operating System and version: CentOS 7 , Debian 11, ArchLinux/CloudLinux 8

PR #298, opened 2022-12-19
feat: add static compilation flag to release

vjerci:
## Description
This pr makes sure that curio runs when c++ toolchain is not installed on system. 

Previously we were depending on c++ being installed on the system, since some distributions like centos, debian, and archlinux in some images don't come with c++ installed, Curio failed to run on those.

## Checklist

### DIFF
--- a/goreleaser/darwin.yaml
+++ b/goreleaser/darwin.yaml
@@ -18,6 +18,7 @@ builds:
       - arm64
     ldflags:
       - -s -w
+      - -extldflags "-static"
       - -X "github.com/bearer/curio/cmd/curio/build.Version={{.Version}}"
       - -X "github.com/bearer/curio/cmd/curio/build.CommitSHA={{.Commit}}"
     hooks:
--- a/goreleaser/linux.yaml
+++ b/goreleaser/linux.yaml
@@ -15,6 +15,7 @@ builds:
       - amd64
     ldflags:
       - -s -w
+      - -extldflags "-static"
       - -X "github.com/bearer/curio/cmd/curio/build.Version={{.Version}}"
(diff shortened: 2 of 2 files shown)


cfabianski (dismissed):

cfabianski:
https://github.com/Bearer/curio/issues/296

goreleaser/darwin.yaml
```diff
@@ -18,6 +18,7 @@ builds:
       - arm64
     ldflags:
       - -s -w
+      - -extldflags "-static"
```
cfabianski, line 21: Should we do it only for Linux? Looks like it’s only an issue there

spdawson (commented):
I was looking at an alternative approach, using nFPM to build Linux packages that declare their dependencies. The static linking approach taken here is probably better, but I'm curious as to what impact this has on the size of the `curio` binary.  It looks like we are currently around 11-12MB; does the static linking dramatically increase this?

goreleaser/darwin.yaml
vjerci: I don't think it can hurt to keep the builds consistent.

vjerci:
> I was looking at an alternative approach, using nFPM to build Linux packages that declare their dependencies. The static linking approach taken here is probably better, but I'm curious as to what impact this has on the size of the `curio` binary. It looks like we are currently around 11-12MB; does the static linking dramatically increase this?

I've compared the binary itself not the tar (tar is 11-12MB) while binary unpacked is 50-60MB The difference between static and non static is in 0.2MB which isn't signficant.

spdawson (approved):

spdawson:
Closes #296

PR #302, opened 2022-12-20
feat: Use encrypted_ column prefix to mark SQL detections encrypted

### LINKED ISSUE #283: Allow `encrypted_` prefix to be marked as encrypted in SQL detections
[`detect_encrypted_ruby_class_properties`](https://github.com/Bearer/curio/blob/fa16564adaea69ab8b8f658dc83bce6f793be150/pkg/commands/process/settings/custom_detector.yml#L214-L225)  is the verifier we use to mark a SQL detection as `encrypted`.

We could add this new processor [here](https://github.com/Bearer/curio/blob/96cb240f7cf6fc0def8447d9faa61a90222fc98c/pkg/report/output/dataflow/datatypes/get_extras.go#L187-L188) that would automatically flag the fields that are prefixed by `encrypted_` as encrypted.

```rego
package bearer.db_encrypted

 import future.keywords

 default encrypted := false

 encrypted = true {
     startswith(lower(input.target.value.field_name),  "encrypted_")
 }

 verified_by := [
     {
         "detector": "db_encrypted",
     }

```

spdawson:
## Description

Closes #283

## Checklist

### DIFF
--- a/integration/flags/.snapshots/TestReportFlags-report-dataflow-db-encrypted
+++ b/integration/flags/.snapshots/TestReportFlags-report-dataflow-db-encrypted
@@ -1,0 +1,149 @@
+data_types:
+    - name: Country
+      detectors:
+        - name: detect_sql_create_public_table
+          locations:
+            - filename: integration/flags/testdata/db_encrypted/structure.sql
+              line_number: 8
+              stored: true
+              parent:
+                line_number: 1
+                content: |-
+                    CREATE TABLE public.Users (
+                      id bigint NOT NULL,
+                      first_name character varying NOT NULL,
+                      last_name character varying NOT NULL,
+                      email character varying NOT NULL,
+                      tanker_encrypted_date_of_birth character varying NOT NULL,
+                      city character varying NOT NULL,
+                      country character varying NOT NULL,
+                      encrypted_gender character varying,
+                    )
+    - name: Date of birth
+      detectors:
+        - name: detect_sql_create_public_table
+          locations:
+            - filename: integration/flags/testdata/db_encrypted/structure.sql
+              line_number: 6
+              stored: true
+              parent:
+                line_number: 1
+                content: |-
+                    CREATE TABLE public.Users (
+                      id bigint NOT NULL,
+                      first_name character varying NOT NULL,
+                      last_name character varying NOT NULL,
+                      email character varying NOT NULL,
+                      tanker_encrypted_date_of_birth character varying NOT NULL,
+                      city character varying NOT NULL,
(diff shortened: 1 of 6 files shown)


pkg/commands/process/settings/processors/db_encrypted.rego
```diff
@@ -0,0 +1,19 @@
+package bearer.db_encrypted
+
+import future.keywords
+
+encrypted contains detection if {
+	detection := input.target_detections[_]
```
cfabianski, line 14: Should we limit it to `SQL` (includes `create_table_sql` custom detection) or `schema_rb`?

integration/flags/.snapshots/TestReportFlags-report-dataflow-db-encrypted
```diff
@@ -0,0 +1,151 @@
...
+                    )
+    - name: Gender
+      detectors:
+        - name: detect_sql_create_public_table
+          locations:
+            - filename: integration/flags/testdata/db_encrypted/structure.sql
+              line_number: 9
+              encrypted: true
+              verified_by:
+                - detector: detect_sql_create_public_table
```
cfabianski: This doesn't seem correct

pkg/commands/process/settings/processors/db_encrypted.rego
```diff
@@ -0,0 +1,46 @@
...
+
+	detection.detector_type in ["schema_rb", "detect_sql_create_public_table"]
+
+	verifications := [verification |
+		encrypted_detection := sql_encrypted[_]
+		detection.value.object_name == encrypted_detection.value.object_name
+		detection.value.field_name == encrypted_detection.value.field_name
+
+		verification := {
+			"detector": encrypted_detection.detector_type,
```
cfabianski: That's probably why the detector reported in the verified_by isn't correct :)

pkg/commands/process/settings/processors/db_encrypted.rego
```diff
@@ -0,0 +1,46 @@
...
+			"filename": encrypted_detection.source.filename,
+			"line_number": encrypted_detection.source.line_number,
```
cfabianski: Don't think that's required but happy to leave it if you think that's useful

cfabianski (commented):
Couple of comments but looks very close

pkg/commands/process/settings/processors/db_encrypted.rego
spdawson: Will get rid of this

cfabianski (approved):

Issue #281, opened 2022-12-16
Rework config file


cfabianski:
The current config file is becoming way too big. It contains `detectors` and `policies` and its content.
We need to keep this folder as slim as possible

Let's add external folders for detectors and policies 

**current**

```yaml
policy:
    only-policy: []
    skip-policy: []

report:
    format: ""
    output: ""
    report: policies

scan:
    context: ""
    debug: false
    disable-domain-resolution: true
    domain-resolution-timeout: 3s
    force: false
    internal-domains: []
    custom_detector: # <--- to be removed
      # Long content
    policies: # <--- to be removed
      # Long content
```

**after**

```yaml
policy:
    only-policy: []
    skip-policy: []

detector:
    only-detector: [] # <--- to add
    skip-detector: [] # <--- to add
  
report:
    format: ""
    output: ""
    report: policies

scan:
    context: ""
    debug: false
    disable-domain-resolution: true
    domain-resolution-timeout: 3s
    force: false
    internal-domains: []
    external-detector-dir: [] #  <--- to add
    external-policy-dir: [] # <--- to add
```

PR #322, opened 2022-12-23
feat: add dirs for loading custom detectors and policies

vjerci:
## Description
This pr reworks config init in a way that we remove custom detector and policy blobs.
Instead 2 new flags are being added, `skip-detector`, `only-detector` which can be used to disable certain detectors.

Our custom detectors are also reworked, instead of having single .yml config blob we now support having per file basis, which should eventually be reworked in supporting directories, when directory structure is fully defined.

Support for external detector and external policy is added. You can specify dirs which we now scan for yaml files to load user defined custom detectors, and user defined custom policies. Modules for custom policies are searched in a relative path from custom policy definition file.

Last but not least, scan caching is improved in a way that we now take into consideration which custom detectors are enabled and their content.

## Closes
https://github.com/Bearer/curio/issues/281

## Checklist


markmichon:
Regarding custom detectors/policies: Flags make sense, but the rest assumes knowledge about how detectors and policies are written. So it’s clear how to import them, but not how to write them in the first place.

cfabianski (approved):

PR #326, opened 2023-01-04
fix: Do not report leakage policy breaches for unique identifiers

### LINKED ISSUE #318: Rails cookies and session custom detections report Unique Identifiers as policy breaches
## Description

The Rails cookies and session custom detectors currently return unique identifiers (`author.user_id`, for example) as policy breaches.

## Expected Behavior

Unique Identifiers should be regarded as safe, and should not be reported as policy breaches.

## Actual Behavior

Unique Identifiers are reported as policy breaches.

## Possible Fix

Should we be triggering the policies at all for Unique Identifiers? Or should these just be regarded as not being risks?

## Your Environment

- Version used: v0.19.0
- Target codebase stack (e.g. Ruby): Ruby
- Link to your project or code sample:

spdawson:
## Description

The Rails cookies and session custom detectors currently return unique identifiers (`author.user_id`, for example) as policy breaches.

## Related

Closes #318

## Checklist


cfabianski (commented):
I don't understand why the snapshots have extra items for Unique Identifier while the gist of this PR is to not have them in the first place. Could you explain please?

spdawson:
> I don't understand why the snapshots have extra items for Unique Identifier while the gist of this PR is to not have them in the first place. Could you explain please?

The new entries are for the _detections_ of the unique identifiers; I thought the idea here was to remove unique identifiers from policy breaches, rather than from the custom detections themselves?

cfabianski (approved):

PR #372, opened 2023-01-10
feat(recipes): add exclusion patterns when wildcard is used

### LINKED ISSUE #370: Do not classify `ajax.googleapis.com` as Google API
## Description

[Google Cloud APIs](https://github.com/Bearer/curio/blob/main/pkg/classification/db/recipes/google_cloud_apis.json#L6) are matching with a wildcard `*.googleapis.com` but `ajax.googleapis.com` shouldn't be considered.

## Expected Behavior

`ajax.googleapis.com` should not be classified and present in the component section

## Actual Behavior

`ajax.googleapis.com` is classified as a Google API and is present in the component section

cfabianski:
## Description

closes #370 

## Checklist

### DIFF
--- a/pkg/classification/interfaces/interfaces.go
+++ b/pkg/classification/interfaces/interfaces.go
@@ -58,6 +58,7 @@ type RecipeURLMatch struct {
 	RecipeUUID       string
 	RecipeName       string
 	RecipeSubType    string
+	ExcludedURL      bool
 }
 
 var ErrInvalidRecipes = errors.New("invalid interface recipe")
@@ -94,14 +95,14 @@ func New(config Config) (*Classifier, error) {
 			preparedRecipe.URLS = append(preparedRecipe.URLS, preparedRecipeURL)
 		}
 
-		for _, ignoredRecipeURL := range recipe.ExcludeURLS {
-			regexpMatcher, err := url.PrepareRegexpMatcher(ignoredRecipeURL)
+		for _, excludedRecipeURL := range recipe.ExcludeURLS {
+			regexpMatcher, err := url.PrepareRegexpMatcher(excludedRecipeURL)
 			if err != nil {
 				return nil, ErrInvalidRecipes
 			}
 
 			preparedRecipeURL := RecipeURL{
-				URL:           ignoredRecipeURL,
+				URL:           excludedRecipeURL,
 				RegexpMatcher: regexpMatcher,
 			}
 			preparedRecipe.ExcludeURLS = append(preparedRecipe.ExcludeURLS, preparedRecipeURL)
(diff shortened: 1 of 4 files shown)


pkg/classification/interfaces/interfaces.go
```diff
@@ -228,15 +254,29 @@ func (classifier *Classifier) Classify(data detections.Detection) (*ClassifiedIn
 	}, nil
 }
 
-func (classifier *Classifier) FindMatchingRecipeUrl(detectionURL string) (*RecipeURLMatch, error) {
+func (classifier *Classifier) FindMatchingRecipeUrl(detectionURL string) (*RecipeURLMatch, *classify.ClassificationDecision, error) {
```
elsapet: Ideally this func wouldn't care about classification and would only care about matching URLs. 
Suggestion to add an ignored or excluded `Boolean` to `RecipeURLMatch`, which we can then use to classify the URL in the calling function above, and simplify the concerns of this func. What do you think?

pkg/classification/interfaces/interfaces.go
```diff
@@ -92,6 +93,20 @@ func New(config Config) (*Classifier, error) {
 			}
 			preparedRecipe.URLS = append(preparedRecipe.URLS, preparedRecipeURL)
 		}
+
+		for _, ignoredRecipeURL := range recipe.ExcludeURLS {
```
elsapet: I think we should pick either ignored or excluded and not mix the two terms.
```suggestion
		for _, excludedRecipeURL := range recipe.ExcludeURLS {
```

elsapet (commented):
Overall looks good but a small suggestion to keep things simple; happy to discuss!

cfabianski:
@elsapet I agree. Updated the PR following your recommendations. Let me know what you think :)

pkg/classification/interfaces/interfaces.go
```diff
@@ -233,6 +263,19 @@ func (classifier *Classifier) FindMatchingRecipeUrl(detectionURL string) (*Recip
...
 	for _, recipe := range classifier.Recipes {
+		for _, recipeURL := range recipe.ExcludeURLS {
+			match, err := url.Match(detectionURL, recipeURL.RegexpMatcher)
+			if err != nil {
+				return nil, err
+			}
+
+			if match != "" {
+				return &RecipeURLMatch{
+					ExcludedURL: true,
```
elsapet, line 274: Since we have this info, we could include it (may be useful for debugging). I'll leave it up to you!

```suggestion
					ExcludedURL: true,
					DetectionURLPart: match,
					RecipeURL:        recipeURL.URL,
					RecipeUUID:       recipe.UUID,
					RecipeName:       recipe.Name,
					RecipeSubType:    recipe.SubType,
```

elsapet (approved):
😍 👏

pkg/classification/interfaces/interfaces.go
cfabianski: Not sure yet 
I'll merge it like this let's see how useful it is :)

PR #374, opened 2023-01-10
spike: refactor policy and detector directory structure

### LINKED ISSUE #336: Revisit structure for detectors/policies
## Description

Curio works with detectors and policies. Detectors are there to enrich the dataflow report, policies are there to look at the dataflow report and apply policies based on the output.

For example:
- a detector would be located in `/pkg/commands/process/settings/custom_detectors/` (e.g. `/pkg/commands/process/settings/custom_detectors/ruby_loggers.yml`)
- a policy would be located in `pkg/commands/process/settings/policies/` (e.g. `/pkg/commands/process/settings/policies/leakage.rego` and an entry inside `/pkg/commands/process/settings/policies.yml`)

Whenever we implement a new detection, we either rely on an existing policy (e.g. leakage applies to both cookies and session), either we enrich the collection of policies.

It might be a bit difficult for a new comer to know where to start especially as we introduce the remediation layer on top.

## Proposal

We could have a structure like this

```bash
/policies/ruby/shared/shared.rego
/policies/ruby/CR-001/detectors/loggers.yml
/policies/ruby/CR-001/leakage.rego
/policies/ruby/CR-001/rule.yml
/policies/ruby/CR-001/README.md
```

We could end up repeating the policies (like `leakage.rego` would be similar for both session and cookies) but that would definitely help any new comer to understand where to start and to contribute.

elsapet:
## Description

closes #336 

Aim: make it easier for newcomers to understand the policy/detector folder structure, so that it's easier for them to contribute. 

I've spiked the following structure; namely, a folder for each policy, with a sub folder for each language that contains a single detector.yml file. 

- `policies/<POLICY_ID>/<POLICY>.rego`
- `policies/<POLICY_ID>/README.md`
- `policies/<POLICY_ID>/rule.yml`
- `policies/<POLICY_ID>/<LANGUAGE>/detector.yml`

e.g. with the logger leaks policy, we have the following: 

- `policies/CR-001/leakage.rego`
- `policies/CR-001/README.md`
- `policies/CR-001/rule.yml`
- `policies/CR-001/ruby/detector.yml`

In addition, I propose some changes to policy IDs and display IDs; namely

- Merge ID and display ID so a policy just has 1 ID e.g. CR-001
- Add the corresponding policy ID to its rule and use this instead of the rule filename
- Remove the 'id' attribute from all YAML files and rely on the folder name (avoid duplication here)

Questions: 

1. Would we want to handle more than 1 x detector per policy per language? 
2. Are we happy to merge display ID and ID - or do we want to retain an internal ID e.g. some UUID that we can change without impact to any documentation or links?
3. Would we want a more generic name for the REGO files, e.g. `CR-001/policy.rego` instead of `CR-001/leakage.rego`? With the proposed folder structure this would now be possible.

Still TODO (i.e. not covered by this spike)

- Fix / test `--only-policy` and `--skip-policy` flags as well as `--only-detector` and `--skip-detector` flags
- Test with multiple languages
- Remove 'language' attribute from YAML file and calculate this from folder name

## Checklist


cfabianski:
@markmichon What do you think?

elsapet:
@markmichon @cfabianski I've spiked the folder structure for 2 x policies, CR-001 and CR-003. To avoid confusion and to make the spike proposal clearer, I've temporarily removed the unused files for all other policies in this commit: https://github.com/Bearer/curio/pull/374/commits/8e64860acd6a4a576e571c679baa44c4bd1e4a3e

Once we're happy with the folder structure, I'll revert this commit and re-introduce the other policies using the new folder structure.

pkg/commands/process/settings/policies/CR-001/rule.yml
```diff
@@ -0,0 +1,9 @@
+description: "Leaking sensitive data to loggers is a common cause of data leaks and can lead to data breaches. This policy looks for instances of sensitive data sent to loggers."
```
markmichon, line 1: @elsapet I wonder if `description` is still needed in `rule.yml`? I think currently it's only used in the docs, but the readme could take care of that.

elsapet: @markmichon good shout - we can keep it just in the readme 👍

markmichon:
The structure within the parent folder makes sense to me. I suspect there's no way around living so deep in `.../process/settings/`? 

Beyond that, it this should work well for docs + the readme inclusion may resolve some implementation concerns I've been working on 👍

gmontard:
@elsapet 

I don't know if we should try to retain the same policies (_and associated files besides detector_) across languages, as proposed in the current directory structure. I guess it's hard to say for now how much those will be the same or not, and therefore if it's worth the common structure or in the contrary it will be more painful. Since we may not know until we expand languages, one way or the other is probably OK for now as long as we keep aware of this potential change.

If we stick to the current proposal, there is still the README that probably need to be "per" language, as we provide code examples, explainers and also very soon remediation strategies - all of which are language specific. @markmichon do you second this?

cc/ @cfabianski

cfabianski:
I think this structure works in some cases, but doesn't in some others (when we shared detections particularly)

For instance:
- How do we structure it for policies that only gets triggered if we detect DataTypes at the language level?
- How do we structure it when we have a detections (or policy) that is shared?
- How do we structure it when detection uses some other detection (e.g. `verifier`)

So the structure could become

```bash
# ... Shared detectors per languages
policies/shared/detectors/<LANGUAGE>/shared_detection_1.yml
policies/shared/detectors/<LANGUAGE>/shared_detection_2.yml
# ... Shared modules used by policies
policies/shared/common.rego
# ... Policy with a detector and a verifier
policies/CR-000/policy.rego
policies/CR-000/README.md
policies/CR-000/rule.yml
policies/CR-000/ruby/detector.yml
policies/CR-000/ruby/verifier.yml
# ... Policy consuming only shared detections
policies/CR-001/policy.rego
policies/CR-001/README.md
policies/CR-001/rule.yml
```

What do you think?

Also, @didroe I'd like to get your thoughts on this since you are working on the architecture which might have an impact on that.

markmichon:
@gmontard Per #336, this proposal of ID/lang/filename is partially my doing. I'm of the opinion that a policy (number, name, etc) as a rule should not be tied to a language. For example: CR-001 will exist in (presumably) every language, so repeating its name, description, rationale, etc seems odd. 

The distinction that logger leaks (for example) differ between languages is an implementation one, not an end user one. Would a security team write individual policies for each language, or would they say "don't leak data to loggers"? This feels like we're working up from detectors/classifications rather than down from policies.

For remediations, I think they can all live in the readme without causing much confusion. If the time comes where each policy is 10+ languages, we'll likely outgrow the current docs anyway.

Either way, we can make the docs work with any structure/format, but it probably doesn't make sense to duplicate the common things if it can be avoided.

cfabianski:
Following some discussions, I think a good structure would look like this

```bash
# ... Shared modules used by policies
policies/shared/common.rego
# ... Policy with a detector and a verifier
policies/CR-000/policy.rego
policies/CR-000/README.md
policies/CR-000/rule.yml
policies/CR-000/detectors.yml # contains all the detections associated to this policy (verifier, ...) 
policies/CR-000/testdata/... # contains all the test data 
policies/CR-000/policy_test.go # run the test on that specific policy
```

Several things:

- We would also get rid of `skip-detector` and `only-detector` (but _maybe_ introduce later a `skip-language` and `only-language` -> to be confirmed)
  - Too confusing otherwise
- The `skip-policy` would disable the detectors associated to it
- We don't have a Policy shared across several languages. The logger leak for instance would require to write a new folder with its policy in it
- We can reference the detections name within the policy so that it is restrictive (right now it can be generic)
- We only have shared components for policies


[EDIT] In the case of `weak library detected` and `weak library used to encrypt a password detected`, I think those 2 policies should be merged and the severity would be higher if we detect password.
This would prevent the need for shared detection which is better

gmontard:
It seems you don't provide any nesting per language here, which I don't think is a good idea.

Language is the starting point of writing policies, our engine, and ultimately what user understand of the compatibility with their stack. Contributor will also contribute mostly only on 1 language, not across languages. 

I believe that sharing policies between languages is a very bad idea - and it's not done by others in the field either - we should collocate everything together from a usage standpoint at the smaller possible granularity, so anyone can quickly understand what a policy does in their context (= a language). Also, we shouldn't care about sharing policies IDs, it doesn't matter, and we will ultimately have a lot of differences between policies across all languages (some more, some less, some very different).

Proposal:

```
policies/ruby/CR-XXX/policy.rego
policies/ruby/CR-XXX/detectors.yml
policies/ruby/CR-XXX/policy.rego
policies/ruby/CR-XXX/README.MD
policies/ruby/CR-XXX/testdata/...
```

And a:
```
policies/shared/detectors-x.yml # Mostly for our master detectors but where we don't anticipate users to contribute to much.
```

The "language" folder should directly map the languages we officially support "ruby", "Javascript" etc., and not include any "framework" which is specific code for a language and as a result should live inside the associated policy detector file if there are any.

cfabianski:
Which form would we rather have?

1️⃣ 

```bash
policies/ruby/CR-XXX/policy.rego
policies/ruby/CR-XXX/detectors.yml
policies/ruby/CR-XXX/policy.rego
policies/ruby/CR-XXX/README.MD
policies/ruby/CR-XXX/testdata/...
policies/rails/CR-XXX/policy.rego
policies/rails/CR-XXX/detectors.yml
policies/rails/CR-XXX/policy.rego
policies/rails/CR-XXX/README.MD
policies/rails/CR-XXX/testdata/...
```

2️⃣
 
```bash
policies/ruby/CR-XXX/policy.rego
policies/ruby/CR-XXX/detectors.yml
policies/ruby/CR-XXX/policy.rego
policies/ruby/CR-XXX/README.MD
policies/ruby/CR-XXX/testdata/...
policies/ruby/rails/CR-XXX/policy.rego
policies/ruby/rails/CR-XXX/detectors.yml
policies/ruby/rails/CR-XXX/policy.rego
policies/ruby/rails/CR-XXX/README.MD
policies/ruby/rails/CR-XXX/testdata/...
```

Let's try to avoid shared detectors but we can come back to it if it becomes an actual usecase. For now, I would merge them and add some additional context to the policy. 
The one use case I have in mind is `bad encryption library` and `bad encryption library used for password` and I think they should be merged. If we only need it for one for now, then it's probably a bad idea to make it part of the design (which makes it more complex IMO)

elsapet:
For clarification (following discussion with @cfabianski), the `rails` framework detectors would be a combination of ruby detection and sql detection

If we do want to include frameworks, I vote for 1️⃣ to avoid the complexity of dealing with the nested structure.

elsapet:
FYI I've started a spike to group by language. It's here in case folk want to have a look at it: https://github.com/Bearer/curio/pull/384

Tomorrow I'll dig a bit deeper and consider a policy with verifier.

cfabianski:
Closing for now

PR #452, opened 2023-01-26
feat: add javascript support

### LINKED ISSUE #431: Add General JS Support
Add JS Support using the new Architecture so that it allows to write custom rules 
- [x] extrapolate common types from ruby
- [x] Implement js language
- [x] Add support for testing compositions
- [x] support object properties `user.name`
- [x] support objects `{ user: name }`
- [x] support classes
   - [x] support constructor signature
   - [x] support method names
- [x] support strings

cfabianski:
## Description

Closes #431 

## Checklist


new/detector/composition/javascript/testdata/testcases/class/constructor.js
```diff
@@ -0,0 +1,3 @@
+class User {
+	constructor(name, city) {}
```
cfabianski, line 2: Should probably support
```javascript
function Person () {
    this.name = 'Sam'
}
```

new/detector/composition/javascript/.snapshots/TestJavascriptClass-method.json
```diff
@@ -0,0 +1,60 @@
+[
```
vjerci: I don't know how to add rule which test for methods

new/detector/composition/javascript/testdata/testcases/class/constructor.js
vjerci: Nice catch.
Would wait for @didroe  on this one as I don't know how to write rule which triggers on `function` node,
I had troubles with this tests case, same troubles we debugged for class. method nodes never get triggered by our framework. 
You'll notice in snapshot that there is no property name or property city on user object, even though tree sitter query should work.

pkg/commands/process/settings/rules/javascript/lang/logger.yml
```diff
@@ -0,0 +1,102 @@
...
+      - variable: DATA_TYPE
+        detection: datatype
+      - variable: METHOD
+        values:
+          - error
+          - log
+          - table
+          - info
+          - warn
+  # fixme: bring elipsis support in javascript
```
didroe: Are we going to deal with this in a separate PR?

new/detector/types/types.go
```diff
@@ -33,6 +33,12 @@ type Detector interface {
 	Close()
 }
 
+type CompositionDetection struct {
```
didroe: I don't think this is used is it?

new/detector/composition/javascript/javascript.go
```diff
@@ -0,0 +1,153 @@
...
+)
+
+type Composition struct {
+	customDetectorTypes []string
+	detectorSet         detectortypes.DetectorSet
+	lang                languagetypes.Language
+	closers             []func()
+}
+
+func New(rules map[string]*settings.Rule, classifier *classification.Classifier) (detectortypes.Composition, error) {
```
didroe, line 35: For a follow-up PR: I think we could share most of the logic here with Ruby. I also think there's opportunity to move most of this to other existing places, and then maybe "composition" mostly/completely disappears. Let's discuss it at some point.

new/detector/implementation/javascript/string/string.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func handleTemplateString(node *tree.Node, evaluator types.Evaluator) ([]interface{}, error) {
+	text := ""
+
+	err := node.EachPart(func(partText string) error {
+		text += partText
+		return nil
+	}, func(child *tree.Node) error {
+		text += "*"
```
didroe, line 76: When the child is `template_substitution`, I think we should take it's named child (`child.Child(1)`) and use similar logic to the concatenation case: 
https://github.com/Bearer/curio/pull/452/files#diff-7bec69bed8205114a204bc68a1a8a822867aa06d1ca446b3f24276bfb5cccd00R51-R63

new/detector/implementation/javascript/property/property.go
```diff
@@ -0,0 +1,120 @@
...
+	result, err := detector.pairQuery.MatchOnceAt(node)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(result) != 0 {
+		key := result["key"]
+		// { user: "admin_user"} || {"user": "admin_user"}
+		if key.Type() == "property_identifier" || key.Type() == "string" {
+			return []interface{}{generictypes.Property{Name: result["key"].Content()}}, nil
```
didroe, line 74: Do we need to strip quotes for the string case?

new/detector/implementation/javascript/property/property.go
```diff
@@ -0,0 +1,120 @@
...
+	node *tree.Node,
+	evaluator types.Evaluator,
+) ([]interface{}, error) {
+	// run method query
+	result, err := detector.methodNameQuery.MatchOnceAt(node)
+	if err != nil || len(result) == 0 {
+		return nil, err
+	}
+
+	log.Debug().Msgf("")
```
didroe, line 101:
```suggestion
```

new/detector/implementation/javascript/property/property.go
```diff
@@ -0,0 +1,120 @@
...
+
+		return properties, nil
+	}
+
+	return []interface{}{generictypes.Property{Name: result["name"].Content()}}, nil
+
+}
+
+func (detector *propertyDetector) Close() {
+	detector.pairQuery.Close()
```
didroe, line 126: We need to clean up the other queries here

new/detector/implementation/javascript/object/object.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	// { user: <object> }
+	parentPairQuery, err := lang.CompileQuery(`(pair key: (_) @key value: (_) @value) @root`)
+	if err != nil {
+		return nil, fmt.Errorf("error compiling parent pair query: %s", err)
+	}
+
+	// class User
+	// end
```
didroe, line 56:
```suggestion
	// class User {
	// }
```

new/detector/implementation/javascript/object/object.go
```diff
@@ -0,0 +1,265 @@
...
+
+	detections, err = detector.getProperties(node, evaluator)
+	if len(detections) != 0 || err != nil {
+		return detections, err
+	}
+
+	return detector.nameParentPairObject(node, evaluator)
+}
+
+func (detector *objectDetector) getobject(
```
didroe, line 121:
```suggestion
func (detector *objectDetector) getObject(
```

new/language/implementation/javascript/javascript.go
```diff
@@ -0,0 +1,240 @@
...
+
+var (
+	variableLookupParents = []string{"pair", "arguments", "binary_expression"}
+
+	anonymousPatternNodeParentTypes = []string{}
+	patternMatchNodeContainerTypes  = []string{}
+
+	// $<name:type> or $<name:type1|type2> or $<name>
+	patternQueryVariableRegex = regexp.MustCompile(`\$<(?P<name>[^>:!]+)(?::(?P<types>[^>]+))?>`)
+	allowedPatternQueryTypes  = []string{"identifier", "property_identifier", "_", "member_expression"}
```
didroe, line 29: Do we need `member_expression` here? I think it's better not to, because of the way things nest. eg. a user might like to do `x.$<VAR>` where `VAR` is filtered on `a.b` but it won't actually match `x.a.b` as it's nested like `(x.a).b` in the AST.

new/language/implementation/javascript/javascript.go
```diff
@@ -0,0 +1,240 @@
...
+			}
+		case "identifier":
+			parent := node.Parent()
+			if parent != nil && slice.Contains(variableLookupParents, parent.Type()) {
+				scopedNode := scope[node.Content()]
+				if scopedNode != nil {
+					node.UnifyWith(scopedNode)
+				}
+			}
+		case "property_identifier":
```
didroe, line 95: I'm not sure we want this case. I think property identifier is just a key name isn't it? ie. not a variable reference.

new/language/implementation/javascript/javascript.go
```diff
@@ -0,0 +1,240 @@
...
+					node.UnifyWith(scopedNode)
+				}
+			}
+		}
+
+		return visitChildren()
+	})
+}
+
+// TODO: See if anything needs to be added here
```
didroe, line 109: LGTM
```suggestion
```

new/language/implementation/javascript/javascript.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	if len(matches) > 1 {
+		return "", 0, errors.New("pattern must only contain a single match node")
+	}
+
+	match := matches[0]
+	return string(inputBytes[0:match[0]]) + string(inputBytes[match[1]:]), match[0], nil
+}
+
```
didroe, line 165: This was replaced by `FindPatternMatchNode`
```suggestion
```

new/language/implementation/javascript/javascript.go
```diff
@@ -0,0 +1,240 @@
...
+// TODO: See if anything needs to be added here
+func produceDummyValue(i int, nodeType string) string {
+	switch nodeType {
+	case "identifier", "call":
+		return "curioVar" + fmt.Sprint(i)
+	default:
+		return "CurioVar" + fmt.Sprint(i)
+	}
```
didroe, line 173: You might get away with this for JS:
```suggestion
	return "curioVar" + fmt.Sprint(i)
```

new/language/implementation/javascript/javascript.go
```diff
@@ -0,0 +1,240 @@
...
+	"github.com/bearer/curio/pkg/util/regex"
+
+	patternquerytypes "github.com/bearer/curio/new/language/patternquery/types"
+	sitter "github.com/smacker/go-tree-sitter"
+)
+
+var (
+	variableLookupParents = []string{"pair", "arguments", "binary_expression"}
+
+	anonymousPatternNodeParentTypes = []string{}
```
didroe, line 24:
```suggestion
	anonymousPatternNodeParentTypes = []string{"binary_expression"}
```

new/language/implementation/javascript/javascript.go
```diff
@@ -0,0 +1,240 @@
...
+func produceDummyValue(i int, nodeType string) string {
+	switch nodeType {
+	case "identifier", "call":
+		return "curioVar" + fmt.Sprint(i)
+	default:
+		return "CurioVar" + fmt.Sprint(i)
+	}
+}
+
+// TODO: See if anything needs to be added here
```
didroe, line 176:
```suggestion
```

new/language/implementation/javascript/javascript.go
```diff
@@ -0,0 +1,240 @@
...
+func (implementation *javascriptImplementation) AnonymousPatternNodeParentTypes() []string {
+	return anonymousPatternNodeParentTypes
+}
+
+// TODO: See if anything needs to be added here
```
didroe, line 181:
```suggestion
```

new/language/implementation/javascript/javascript.go
```diff
@@ -0,0 +1,240 @@
...
+func (implementation *javascriptImplementation) FindPatternMatchNode(input []byte) [][]int {
+	return matchNodeRegex.FindAllIndex(input, -1)
+}
+
+// TODO: See if anything needs to be added here
```
didroe, line 186:
```suggestion
```

new/language/implementation/javascript/javascript.go
```diff
@@ -0,0 +1,240 @@
...
+func (implementation *javascriptImplementation) FindPatternUnanchoredPoints(input []byte) [][]int {
+	return ellipsisRegex.FindAllIndex(input, -1)
+}
+
+// TODO: See if anything needs to be added here
+func (implementation *javascriptImplementation) IsTerminalDetectionNode(node *tree.Node) bool {
+	return false
+}
+
```
didroe, line 195: Replaced by `DescendIntoDetectionNode`
```suggestion
```

new/language/implementation/javascript/javascript.go
```diff
@@ -0,0 +1,240 @@
...
+// TODO: See if anything needs to be added here
```
didroe: Empty list is probably good for JS
```suggestion
```

new/language/tree/node.go
```diff
@@ -67,6 +67,45 @@ func (node *Node) Child(i int) *Node {
 	return node.tree.wrap(node.sitterNode.Child(i))
 }
 
+func (node *Node) EachPart(onText func(text string) error, onChild func(child *Node) error) error {
```
didroe: This might be more descriptive:
```suggestion
func (node *Node) EachContentPart(onText func(text string) error, onChild func(child *Node) error) error {
```

new/language/implementation/javascript/javascript.go
```diff
@@ -0,0 +1,240 @@
...
+func (implementation *javascriptImplementation) PatternMatchNodeContainerTypes() []string {
+	return patternMatchNodeContainerTypes
+}
+
+func (implementation *javascriptImplementation) PatternIsAnchored(node *tree.Node) bool {
+	// FIXME: implement this
```
didroe: Looks like this is done?
```suggestion
```

new/language/implementation/javascript/javascript.go
```diff
@@ -0,0 +1,240 @@
...
+	if parent.Type() == "function_declaration" && !node.Equal(parent.ChildByFieldName("name")) && !node.Equal(parent.ChildByFieldName("parameters")) {
+		return false
+	}
+
+	// method
+	if parent.Type() == "method_definition" && !node.Equal(parent.ChildByFieldName("name")) && !node.Equal(parent.ChildByFieldName("parameters")) {
+		return false
+	}
+
+	return true
```
didroe: I think this should be something like:
```suggestion
	return parent.Type() != "class_body" && parent.Type() != "statement_block"
```

new/language/implementation/javascript/javascript.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// TODO: See if anything needs to be added here
+func (implementation *javascriptImplementation) DescendIntoDetectionNode(node *tree.Node) bool {
+	// parent := node.Parent()
+
+	// if parent != nil && parent.Type() == "member_expression" && node.Equal(parent.ChildByFieldName("object")) {
+	// 	return slice.Contains(passThroughMethods, parent.ChildByFieldName("method").Content())
+	// }
```
didroe: We probably want something like this:
```suggestion
	parent := node.Parent()

	if parent != nil && parent.Type() == "member_expression" && node.Equal(parent.ChildByFieldName("object")) {
	 	return false
	}
```

didroe (commented):
Nice work

new/detector/types/types.go
vjerci: nice catch

new/detector/composition/javascript/javascript.go
vjerci: Looking forward to doing that :)
But it might be out of scope for this pr.

new/detector/implementation/javascript/property/property.go
vjerci: nice catch, I've added code to handle it as well as test case

new/language/implementation/javascript/javascript.go
vjerci: Saw it implemented like this in ruby and thought there is some reason for it.
will change it.

didroe: In Ruby the expressions are directly under the class/block where there can be other things like the class name/block args. But in JS it's nicely in a separate node.

PR #461, opened 2023-01-27
feat: test framework for new detectors

vjerci:
## Description
This pr adds test framework for new detectors allowing us to run unit based snapshot tests on compositions.
Reason fro adding unit tests is because running unit tests, compared to running integration tests is considerably faster as it doesn't require us to spawn additional process.

This pr also extrapolates some of the generic types that will be used in javascript detectors.

## Checklist


new/detector/composition/ruby/ruby.go
```diff
@@ -127,108 +129,116 @@ func (composition *Composition) Close() {
 	}
 }
 
-func (composition *Composition) DetectFromFile(report report.Report, file *file.FileInfo) error {
+func (composition *Composition) DetectFromFile(file *file.FileInfo) ([]compositiontypes.Detection, error) {
```
didroe: Instead of introducing a new detection type. WDTY about returning the original detections here and doing the translation outside of `Composition`? I think the only thing you don't have on the original detection that you'd need is the detector type. But we can add that to the type + all the code that builds them in the detectors*.

\* This boilerplate can probably go away when the current change I'm making is done. Following my change, `MatchNode` will always be the node that is passed to the detector, so detectors could just return an array of `Data`s and the evaluator could wrap them in a `Detection` (adding `MatchNode` and `DetectorType`)

vjerci: About this translation boilerplate: I agree with you and I wanted to do that just that I didn't have time to code it.
Anyways I did it now. Will push changes as soon as I check tests.

cfabianski (approved):

PR #474, opened 2023-01-31
feat: match equivalent ruby syntax

didroe:
## Description

Make the following Ruby syntax equivalent for matching against patterns:
- `key: value` vs `:key => value`
- `call do ... end` vs `call { ... }`

## Checklist


cfabianski (approved):
very nice!

vjerci (approved):

PR #475, opened 2023-01-31
chore: clean up

cfabianski:
## Description

Clean up legacy

- moved sql create table rules to its own `built_in_rules` folder (@markmichon will make the documentation simpler as it was a special snowflake until then)
- remove create view detection (not currently used)
- moved devise patterns into its own file

## Checklist


pkg/commands/process/settings/rules/ruby/third_parties/devise.yml
```diff
@@ -0,0 +1,54 @@
...
+      end
+    filters:
+      - either:
+          - variable: MAX_LENGTH
+            less_than: 35
+          - variable: MIN_LENGTH
+            less_than: 8
+  - pattern: |
+      Devise.setup do |$<CONFIG:identifier>|
+        $<CONFIG>.password_length = $<MIN_LENGTH>..$<MAX_LENGTH>
```
didroe:
```suggestion
        $<!>$<CONFIG>.password_length = $<MIN_LENGTH>..$<MAX_LENGTH>
```

pkg/commands/process/settings/rules/ruby/third_parties/devise.yml
```diff
@@ -0,0 +1,54 @@
...
+      end
+    filters:
+      - variable: MIN_LENGTH
+        less_than: 8
+  - pattern: |
+      Devise.setup do |$<CONFIG:identifier>|
+        $<CONFIG>.password_length = $<LENGTH>
```
didroe:
```suggestion
        $<!>$<CONFIG>.password_length = $<LENGTH>
```

pkg/commands/process/settings/rules/ruby/third_parties/devise.yml
```diff
@@ -0,0 +1,54 @@
...
+      end
+    filters:
+      - variable: LENGTH
+        less_than: 8
+        match_violation: true
```
didroe: This isn't supported any more
```suggestion
```

pkg/commands/process/settings/rules/ruby/third_parties/devise.yml
```diff
@@ -0,0 +1,54 @@
...
+detect_presence: true
```
didroe:
```suggestion
```

pkg/commands/process/settings/settings.go
```diff
@@ -164,6 +170,7 @@ func (rule *Rule) PolicyType() bool {
 
 func FromOptions(opts flag.Options) (Config, error) {
 	policies := DefaultPolicies()
+	builtInRules := BuiltInRules()
```
didroe: Do these need to be completely separate? Or could we just load them from a separate location? ie. add them to `loadRules`

cfabianski: I thought it was the easiest to keep things separated (at least for now) 
See https://github.com/Bearer/curio/pull/475/files#diff-cf923dbfaf4cadc868828b30bfb30d1fd225d3178248d9521dca28be0504446dR37

elsapet:
@cfabianski Thinking about it, I'm wondering if Devise should stay in the `rails/` folder - by their description, they are a Rails (rather than Ruby) solution: https://github.com/heartcombo/devise

pkg/commands/process/settings/settings.go
didroe: Fair enough. Maybe you can re-use pieces of `loadRules`? I'd like to avoid the chance of loading things in a different way if possible.

markmichon (approved):

Issue #398, opened 2023-01-16
Privacy Report


gotbadger:
As the basics technical requirements of each laws are similar in their requirements (maintain an inventory of assets, know what data types are shared with 3rd parties, ensure data is adequately protected), we can make a Privacy Report that will tackle a large range of laws while automating the laborious tasks that legal teams require engineering to perform.

This feature will leverage Curio's detections capabilities and put it in a format that will make exploitation by legal teams easier. Our goal is to reduce the friction between engineering and legal teams. 

## 1. Subjects and Data-types Inventory 

This report will compute together all Subjects and Data Types detected through codebase scan. 

To detect subject, we will use the "object" item when they are classified as a type of person. Our current list of Subjects is : 

{TO-DO: retrieve the list of subjects we currently have embedded in the codebase}

Curio will also report the number of policies tested against the data found in the codebase. It will list how many are failing, their severity, and how many have passed.

#### Output format: 

| Subject | Data Types | Detection Count | Critical Policy Failure | High Policy Failure | Medium Policy Failure | Low Policy Failure | Passed Policy |
| --- | --- | --- | --- | --- | --- | --- | --- |
| Customer | Full Name | 142 | 1 | 3 | 0 | 15 | 184 |
| Customer | E-mail | 84 | 0 | 0 | 4 | 1 | 184 |
| Customer | Address | 42 |  0 | 0 | 4 | 0 | 184 |
| Customer | Password | 21 | 1 | 0 | 0 | 0 | 184 |
| Admin | Full Name | 18 | 0 | 0 | 0 | 1 | 184 |
| Admin | E-mail | 12 | 0 | 0 | 0 | 1 | 184 |

Cardinality: one line per data-type per subject. 

Must-Have output format is CSV. Other formats are nice to have. 

## 2. Third-Parties Data Sharing Inventory

Curio has the ability to detect Third Party usage through codebase inspection. Read more about this here.

For each detected Third-Party, Curio will list all Subjects & Data Types determined to be exchanged through the integration. If Curio isn't capable of determining what Data are exchanged with a Third party, it will report an unknown state. 

Curio will also report the number of policies tested against the data found in the codebase. It will list how many are failing, their severity, and how many have passed.

#### Output file:

| Third Party | Subject | Data Types | Critical Policy Failure | High Policy Failure | Medium Policy Failure | Low Policy Failure | Passed Policy |
| --- | --- | --- | --- | --- | --- | --- | --- |
[...]

PR #478, opened 2023-02-01
feat(privacy report): pass flag option for subject mapping override

elsapet:
## Description
* Allows users to override data subject mapping by providing a file path to their own mapping. This file is then read during the scan process

References #398

## Checklist

- [ ] I've included usage information in the description if CLI behaviour was updated or added.

### DIFF
--- a/docs/_data/curio_scan.yaml
+++ b/docs/_data/curio_scan.yaml
@@ -7,6 +7,9 @@ options:
     - name: context
       usage: |
         Expand context of schema classification e.g., --context=health, to include data types particular to health
+    - name: data-subject-mapping
+      usage: |
+        Override default data subject mapping by providing a path to a custom mapping JSON file
     - name: debug
       default_value: "false"
       usage: Enable debug logs
--- a/integration/flags/.snapshots/TestInitCommand
+++ b/integration/flags/.snapshots/TestInitCommand
@@ -7,6 +7,7 @@ rule:
     skip-rule: []
 scan:
     context: ""
+    data_subject_mapping: ""
     debug: false
     disable-domain-resolution: true
     domain-resolution-timeout: 3s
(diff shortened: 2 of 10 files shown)


docs/_data/curio_scan.yaml
```diff
@@ -7,6 +7,9 @@ options:
     - name: context
       usage: |
         Expand context of schema classification e.g., --context=health, to include data types particular to health
+    - name: data-subject-mapping
+      usage: |
+        Override default data subject mapping by providing a path to a custom mapping JSON file
```
elsapet, line 12: @markmichon I'm not sure on the process of adding new flags, but would like your thoughts on the wording for this one. 

Some background: this flag lets users override the default data subject mapping used for the privacy reports, by providing a path to their own JSON file that follows this format https://github.com/Bearer/curio/blob/main/pkg/classification/db/subject_mapping.json

e.g. 

```
curio scan /example/sentry_ex  --report=privacy --data-subject-mapping=/Users/elsapet/Desktop/example/subject_mapping.json
```

markmichon: The flag text looks fine. I think we're going to need a better format example than that, specifically what is allowed other than `user`. Is this documented internally anywhere or do we validate it?  

And we'll need some extra details on why someone might want to do it once the privacy report is up and running.

markmichon (approved):

docs/_data/curio_scan.yaml
elsapet: Thanks @markmichon 🙏 

> Is this documented internally anywhere or do we validate it?

Right now there's no validation, so you'd be free to put any value/s in place of "User". We also don't validate that you've given us a sensible JSON file, which is maybe something we'd want to add too 🤔

cfabianski (approved):

PR #479, opened 2023-02-01
fix(privacy report): clean up for privacy reports

elsapet:
## Description
* Change `--report=privacy` to the more descriptive `--report=subjects`
* Remove confusing progress bar during third party report generation (we don't run all the rules in this case, so it's not necessary)

## Checklist


elsapet:
Since we are merging the two reports, I shall incorporate these changes into that PR instead

Issue #424, opened 2023-01-20
Let user know in CLI when a rescan is using cached data


b1pb1p:
When a scan is launched on a codebase that has already been scanned, Curio will use cached detections and dataflows. It will rescan only if the git commit has changed. If the user has changed the codebase but didn't make a commit, Curio won't rescan the codebase and won't change its policy test results.

We should let users know that we are using cached results when rescanning and hint to use the --force flag to force a rescan if they didn't commit code changes but still expect a different policy test outcome.

PR #500, opened 2023-02-06
feat: warn when cached data is used

elsapet:
## Description

Adds some warning text to the console output, to tell the user when we're using cached data and to prompt them to use `--force` to force a new scan (if they want)

References #424 

### Screenshots

Screenshot 2023-02-06 at 16 01 03

### Summary 

Screenshot 2023-02-06 at 15 59 18

### Stats

Screenshot 2023-02-06 at 16 56 23

### Privacy

Screenshot 2023-02-06 at 16 56 54

## Checklist

### DIFF
--- a/pkg/commands/artifact/run.go
+++ b/pkg/commands/artifact/run.go
@@ -219,7 +219,7 @@ func Run(ctx context.Context, opts flag.Options, targetKind TargetKind) (err err
 	case TargetFilesystem:
 		if report, err = r.ScanFilesystem(ctx, opts); err != nil {
 			if errors.Is(err, balancer.ErrFileListEmpty) {
-				output.StdOutLogger().Msgf("directory empty: %s", err)
+				outputhandler.StdOutLogger().Msgf("directory empty: %s", err)
 				os.Exit(0)
 				return
 			}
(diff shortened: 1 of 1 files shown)


elsapet:
Draft pending product approval of wording

pkg/commands/artifact/run.go
```diff
@@ -60,12 +60,12 @@ type runner struct {
 }
 
 // NewRunner initializes Runner that provides scanning functionalities.
-func NewRunner(ctx context.Context, scanSettings settings.Config) Runner {
+func NewRunner(ctx context.Context, scanSettings *settings.Config) Runner {
```
elsapet: Not sure if this (editing the config object directly) is best practice or no; open to alternative solutions. Maybe @vjerci or @didroe you have some thoughts on this one 🙏

didroe: WDYT about putting the flag on the `Runner` instead?

elsapet: Thanks for the suggestion @didroe - much better!!

cfabianski (approved):

Issue #366, opened 2023-01-10
Increase detections for third-party loggers


gotbadger:
### Proposal
We can improve our rules by adding some rules for the following monitoring tools, I've also looked at the number of downloads for their gems to give a rough idea of popularity.
- [x] NewRelic [(119M)](https://rubygems.org/gems/newrelic_rpm)
  -  log collection on by default in [latest version](https://docs.newrelic.com/docs/logs/logs-context/configure-logs-context-ruby/)
  -  config options we [could check](https://docs.newrelic.com/docs/apm/agents/ruby-agent/configuration/ruby-agent-configuration/) and [Tracer Options](https://docs.newrelic.com/docs/apm/transactions/transaction-traces/configure-transaction-traces/#config-attributes)
  - [x] Manual [Errors](https://docs.newrelic.com/docs/apm/agents/manage-apm-agents/agent-data/manage-errors-apm-collect-ignore-or-mark-expected/#error-collection) aka notice
- [x] Sentry [(83M) old gem](https://rubygems.org/gems/sentry-raven)
  - Check [Fitering of params](https://docs.sentry.io/platforms/ruby/data-management/sensitive-data/)
  - [x] [Tagging and Tracing](https://docs.sentry.io/platforms/ruby/guides/rails/enriching-events/)
- [x] Rollbar [(44M)](https://rubygems.org/gems/rollbar)
  - [x] [Rollbar.error etc.](https://docs.rollbar.com/docs/plain-ruby)
  - `before_process`
  - `transform`
  - `custom_data_method`
  - `log_payload`?
  - `person_method` (and individual config options)
  - `scrub_fields`

- [x] Datadog [(44M)](https://rubygems.org/gems/ddtrace) - log collection @didroe 
  - Filtering params
  - https://docs.datadoghq.com/logs/log_collection/ruby/
  - [x] Spans - https://docs.datadoghq.com/tracing/guide/add_span_md_and_graph_it/?code-lang=ruby
  - Scrubbing [data](https://docs.datadoghq.com/tracing/configure_data_security/?tab=mongodb#scrub-sensitive-data-from-your-spans)

- [x] Airbrake [(37M)](https://rubygems.org/gems/airbrake)  
  - [x] [Notify](https://docs.honeybadger.io/lib/ruby/getting-started/reporting-errors/#reporting-errors-without-an-exception) @elsapet https://github.com/Bearer/curio/pull/514
[...]

PR #509, opened 2023-02-07
feat: add bugsnag and honeybadger support

cfabianski:
## Description

#366 

## Checklist


pkg/commands/process/settings/rules/ruby/third_parties/bugsnag/testdata/before_bugsnag_notify_add_metadata.rb
```diff
@@ -0,0 +1,15 @@
+class ApplicationController < ActionController::Base
+  before_bugsnag_notify :add_diagnostics_to_bugsnag
```
cfabianski, line 2: Need to allow simple_symbol reconciliation with method name reconciliation (which will be done as a separate PR)

PR #514, opened 2023-02-08
feat(ruby rules): add third-party airbrake rule for notify methods

elsapet:
## Description

Adds new Ruby third-party rule to cover [Airbrake](https://github.com/airbrake/airbrake)'s `notify` methods 

Filtering params still TODO 

Relates to #366 

## Checklist


pkg/commands/process/settings/rules/ruby/third_parties/airbrake.yml
```diff
@@ -0,0 +1,74 @@
...
+          - notify
+          - notify_sync
+          - notify_request
+          - notify_request_sync
+          - notify_query
+          - notify_query_sync
+          - notify_performance_breakdown
+          - notify_performance_breakdown_sync
+          - notify_queue
+          - notify_queue_sync
```
elsapet, line 39: Not sure if we could use an auxiliary rule to avoid the repetition here. I chose not to regardless because I think it keeps the rule easy to understand, but open to update it if folk feel strongly.

pkg/commands/process/settings/rules/ruby/third_parties/airbrake.yml
```diff
@@ -0,0 +1,74 @@
...
+  - pattern: |
+      Airbrake.merge_context($<...>$<DATA_TYPE>$<...>)
+    filters:
+      - variable: DATA_TYPE
+        detection: datatype
+  - pattern: |
+      def to_airbrake
+        $<...>$<DATA_TYPE>$<...>
```
didroe: Because it's in a method body, the `...` is implied. You should only need to use the ellipsis in either an arguments list, or after a method call to specify an optional block.

```suggestion
        $<DATA_TYPE>
```

didroe (approved):

Issue #504, opened 2023-02-06
Write JS Rules


cfabianski:
- [x] Loggers @vjerci 
- [x] Cookies @vjerci 
- [x] Session @vjerci 
- [x] File Generation @cfabianski https://github.com/Bearer/curio/pull/546
- [x] Sentry @cfabianski https://github.com/Bearer/curio/pull/526
- [x] Exception @cfabianski https://github.com/Bearer/curio/pull/540
- [x] Weak Encryption @vjerci 
- [ ] SSL Verification @vjerci (I believe this one isn't feasible to do) 
- [x] JWT @vjerci https://github.com/Bearer/curio/pull/549
- [x] HTTP Get/Post @vjerci https://github.com/Bearer/curio/pull/551
- [x] Google Analytics @vjerci 
     - [x] Google tag manager @vjerci https://github.com/Bearer/curio/pull/554
     - [x] Google analytics frontend (without gtm) @vjerci  https://github.com/Bearer/curio/pull/566
     - [x] Google Analytics react library @vjerci  https://github.com/Bearer/curio/pull/567
- [x] Algolia (1,131K weekly downloads on NPM) @elsapet 
    - https://www.algolia.com/doc/api-client/getting-started/install/javascript/?client=javascript
    - https://www.npmjs.com/package/algoliasearch
- [x] Elasticsearch (771K weekly downloads on NPM) @vjerci https://github.com/Bearer/curio/pull/580
     - https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/introduction.html
     - https://www.npmjs.com/package/@elastic/elasticsearch
- [x] Datadog   @vjerci 
  - [x] hot shots 1.5M weekly downloads @vjerci  https://github.com/Bearer/curio/pull/576
  https://www.npmjs.com/package/hot-shots
  - [x] Browser RUM integration (1 M weekly downloads) @vjerci https://github.com/Bearer/curio/pull/581
     - https://docs.datadoghq.com/integrations/rum_javascript/#pagetitle
     - https://www.npmjs.com/package/@datadog/browser-rum
  - [Data receiving only - not relevant] Typescript API(300K weekly downloads) @vjerci this one is for getting data
     - https://docs.datadoghq.com/api/latest/?code-lang=typescript
     - https://www.npmjs.com/package/@datadog/datadog-api-client
- [x] NewRelic (785K weekly downloads on NPM) @elsapet 
[...]

PR #517, opened 2023-02-08
feat: cookie rule

vjerci:
## Description
This pr adds support for insecure cookies.
It also adds support for unanchored object pairs in javascript as well as substituting template literal variables. 

### Related
https://github.com/Bearer/curio/issues/504

## Checklist

### DIFF
--- a/pkg/commands/process/settings/rules/javascript/lang/logger.yml
+++ b/pkg/commands/process/settings/rules/javascript/lang/logger.yml
@@ -1,4 +1,4 @@
-type: "risk"
+type: risk
 patterns:
   - pattern: |
       $<LOG>.$<METHOD>($<DATA_TYPE>)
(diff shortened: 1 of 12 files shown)


pkg/commands/process/settings/rules/javascript/lang/logger.yml
```diff
@@ -1,4 +1,4 @@
-type: "risk"
+type: risk
```
didroe, line 1: Accidental change? (and following one)

didroe (approved):

pkg/commands/process/settings/rules/javascript/lang/logger.yml
vjerci: it's intentional, `"` are redundant here, @cfabianski and me were removing them on some call so figured it's a standard we want to introduce

PR #520, opened 2023-02-08
feat: add ruby open telemetry rule

elsapet:
## Description
Add Ruby rule for open telemetry tracing

Relates to #366 

## Checklist

### DIFF
--- a/pkg/commands/process/settings/rules/ruby/third_parties/open_telemetry/testdata/datatype_in_span_attributes.rb
+++ b/pkg/commands/process/settings/rules/ruby/third_parties/open_telemetry/testdata/datatype_in_span_attributes.rb
@@ -1,0 +1,21 @@
+# add attributes
+def track_user(user)
+  current_span = OpenTelemetry::Trace.current_span
+
+  current_span.add_attributes({
+    "user.id" => user.id,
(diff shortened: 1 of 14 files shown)


pkg/commands/process/settings/rules/ruby/third_parties/open_telemetry/testdata/datatype_in_span_attributes.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  current_span.add_attributes({
+    "user.id" => user.id,
+    "user.first_name" => user.first_name
+  })
+end
+
+# add attributes at span creation
+Tracer.in_span("data leaking", attributes: { "current_user" => user.email, "date" => DateTime.now }) do |span|
+  puts "in the span block"
```
didroe: I think you could then add more attributes in the block with `span.add_attributes/set_attribute`. Maybe we should add an example for that too?

(also for `span.add_event`)

elsapet: Ah good shout; shall do 🙏

didroe (approved):

PR #526, opened 2023-02-08
feat(rules): add sentry rules

cfabianski:
## Description

Sentry JS rules see #504

## Checklist


pkg/commands/process/settings/rules/javascript/third_parties/sentry.yml
```diff
@@ -0,0 +1,72 @@
+patterns:
+  - pattern: |
+      Sentry.addBreadcrumb($<DATA_TYPE>)
+    filters:
+      - variable: DATA_TYPE
+        detection: datatype
```
elsapet: Can we use a variable here or is it not supported for JS e.g. 
```suggestion
  - pattern: |
      Sentry.$<METHOD>($<DATA_TYPE>)
    filters:
      - variable: DATA_TYPE
        detection: datatype
      - variable: METHOD
         values: 
           - addBreadcrumb
           - captureMessage
           ...
```

cfabianski: We probably can but I keep that for after :)

Issue #459, opened 2023-01-27
Do not send sensitive data to analytic environnements.


gotbadger:
### Proposal
Its easy to send sensitive data to places we should not. This is a rundown of analytic tools that might consume this data. I've ordered the vendors/tools by popularity and then support

- Redshift ([64M](https://rubygems.org/gems/aws-sdk-redshift))
  - Official [ruby gem](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/Redshift.html) 
- [x] Algolia ([36M](https://rubygems.org/gems/algoliasearch)) @didroe 
  - [x] [saving](https://www.algolia.com/doc/api-reference/api-methods/save-objects/#examples)
  - [indexing](https://www.algolia.com/doc/framework-integration/rails/indexing/serializing-models/?client=ruby)
- [x] Elastic ([33M](https://rubygems.org/gems/elasticsearch-rails)) @didroe 
  -   Project contains seperate gems for bulk imports as well as persistance operations and is [documented](https://github.com/elastic/elasticsearch-rails#usage)
  - [x] Manual index
  - Models
  - Persistence
  - [x] Bulk import
- [x] Segment analytics ([19M](https://rubygems.org/gems/analytics-ruby)) @elsapet 
  - [ident & track methods](https://github.com/segmentio/analytics-ruby#usage)
  - https://github.com/Bearer/curio/pull/539

- [x] Google BigQuery ([~9M](https://rubygems.org/gems/google-cloud-bigquery)) @didroe 
  - Well documented, mainly around querying as you might expect but here is the [stuff on inserts](https://cloud.google.com/ruby/docs/reference/google-cloud-bigquery/latest/Google-Cloud-Bigquery-Dataset#Google__Cloud__Bigquery__Dataset_insert_instance_)
- [x] Google Dataflow ([~7k-9k](https://rubygems.org/gems/google-cloud-dataflow)) @elsapet 
  - [Versioned clients](https://cloud.google.com/ruby/docs/reference/google-cloud-dataflow/latest) might make this a bit hard to support well
  - Main page https://github.com/googleapis/google-cloud-ruby/tree/main/google-cloud-dataflow
  - NOT DONE: Protobuf support. Protobuf is used for metadata in some cases. Suggestion to handle as its own case instead. 
- [x] Google Analytics @elsapet 
[...]

PR #539, opened 2023-02-09
feat: add ruby segment rule

elsapet:
## Description

Add rule for Ruby's Segment Analytics library

References #459 

## Checklist


pkg/commands/process/settings/rules/ruby/third_parties/segment/.snapshots/TestRubyThirdPartiesSegmentDataflow-dataflow_ruby_third_parties_segment_ok_no_datatypes.rb
```diff
@@ -0,0 +1,5 @@
+components: []
+
+
+--
+
```
elsapet, line 5: I thought we would have detected the `segment_init` in risks here 🤔 Any thoughts @didroe ?

pkg/commands/process/settings/rules/ruby/third_parties/segment.yml
```diff
@@ -0,0 +1,45 @@
...
+    ## Description
+    Leaking sensitive data to third-party analytics tools is a common cause of data leaks and can lead to data breaches. This rule looks for instances of sensitive data sent to Segment.
+
+    ## Remediations
+
+    When sending data to analytics libraries, ensure all sensitive data is removed.
+
+    ## Resources
+    - [Segment docs](https://segment.com/docs/connections/sources/catalog/libraries/server/ruby/)
+  dsr_id: DSR-1
```
elsapet, line 43: Here I've kept the description and remediation quite the same as the logger copy, but in this case Segment is not a logger, it's an analytics tool. 

The DSR ID is the same, I think. 

Ping @markmichon for any thoughts or suggestions  🙏

markmichon: Thanks, lgtm 🙇

cfabianski (approved):

PR #541, opened 2023-02-09
fix: resolve homedir when loading external rules

### LINKED ISSUE #529: external-rule-dir flag fails when combining = syntax with ~ path
## Description

When loading a custom rule, the `--external-rule-dir` flag fails and throws an error for the following:

```bash
go run cmd/curio/main.go scan ~/_dev/bearer/bear-publishing/ --external-rule-dir=~/_dev/bearer/curio-rules
```
Resulting in this error:
```
Error: error loading external rules from ~/_dev/bearer/curio-rules: stat ~/_dev/bearer/curio-rules/.: no such file or directory
error loading external rules from ~/_dev/bearer/curio-rules: stat ~/_dev/bearer/curio-rules/.: no such file or directory
exit status 1
```

But does not fail for this command:
```bash
go run cmd/curio/main.go scan ~/_dev/bearer/bear-publishing/ --external-rule-dir=/Users/markmichon/_dev/bearer/curio-rules
```

And does not fail for this command:
```bash
go run cmd/curio/main.go scan ~/_dev/bearer/bear-publishing/ --external-rule-dir ~/_dev/bearer/curio-rules
```

This leads me to think the combination of `--external-rule-dir=` syntax (equal syntax) along with the tilde(~) is the root of the problem. Not sure if this is a limitation of = syntax, or a bug.

## Your Environment

- Version used: Main
- Target codebase stack (e.g. Ruby): ruby
- Operating System and version: macos venture 13.1

gotbadger:
## Description
- When using `--external-rule-dir=~/my-rule-dir` the path is not expanded by the shell and is passed literally to the application. This PR expands out `~` to give the full path again in this situation
- Added debug for loading external rules so users can check they are being loaded at application startup

## Related
- Close #529 

## Checklist

### DIFF
--- a/pkg/commands/process/settings/rules.go
+++ b/pkg/commands/process/settings/rules.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/bearer/curio/pkg/flag"
+	"github.com/rs/zerolog/log"
 	"golang.org/x/exp/maps"
 	"gopkg.in/yaml.v3"
 )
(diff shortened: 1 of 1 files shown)


didroe (commented):
I'm not sure about this change. I feel like we should let the shell deal with expanding the path.

eg. you could just call it like this and it would work:

```sh
$ curio scan --external-rule-dir ~/my-rule-dir ...
```

cfabianski (approved):

PR #550, opened 2023-02-10
docs(readme): update Debian installation step

### LINKED ISSUE #547: Documentation: Not able to install curio on Ubuntu with the instructions given
Installation fails when following instructions found in the readme (https://github.com/Bearer/curio#debianubuntu) on Ubuntu. 

The following line
```
echo deb [trusted=yes] https://apt.fury.io/bearer/ / | sudo tee -a /etc/apt/sources.list.d/fury.list
```

results in 
image

which leads to package not found when trying to install. 

As you can see from the screenshot, I am using `zsh` and it might be the reason why this fails. 

Hope that helps.

cfabianski:
## Description

Following [gemfury documentation](https://gemfury.com/help/apt-repository/), that's how it should be

## Related

- #547

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ brew install Bearer/curio/curio
 
 ```shell
 $ sudo apt-get install apt-transport-https
-$ echo deb [trusted=yes] https://apt.fury.io/bearer/ / | sudo tee -a /etc/apt/sources.list.d/fury.list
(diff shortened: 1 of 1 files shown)


gotbadger:
@cfabianski ill just verify this works and if its good then we can merge it

gotbadger (approved):

PR #565, opened 2023-02-13
fix: ruby file generation

didroe:
## Description

Fix Ruby file generation rule:
- `CSV.generate` has no arguments and the rule was requiring them
- Allow variable lookup in `binary` expression - this allows `csv << row` to match correctly

## Checklist


elsapet (approved):
🙏 🎉

PR #567, opened 2023-02-13
feat: add javascript react google analytics rule

vjerci:
## Description
This rule searches for sensitive data being sent to google analytics using popular ReactGAlibrary. 

pattern: 
```
patterns:
  - pattern: |
      ReactGA.event($<DATA_TYPE>)
    filters:
      - variable: DATA_TYPE
        detection: datatype
```

## Checklist


pkg/commands/process/settings/rules/javascript/react/google_analytics.yml
```diff
@@ -0,0 +1,31 @@
+patterns:
+  - pattern: |
+      ReactGA.event($<DATA_TYPE>)
```
didroe, line 3: Should this be it's own rule? Or should we add this to `third_parties/google_analytics.yml`?

elsapet (approved):

pkg/commands/process/settings/rules/javascript/react/google_analytics.yml
vjerci: It is a non official libarary, so to me it made sense to add it as separate rule.
However, I don't have strong feelings about it and can see your point of it belonging to google analyitics rule.
Am merging it as is in order to move forward but please let me know if you want it changed.

PR #580, opened 2023-02-15
feat: add elasticsearch

vjerci:
## Description
This pr adds a rule for sensitive data in elasticsearch database

Patterns:
```
  - pattern: |
      $<CLIENT>.index($<...>$<DATA_TYPE>$<...>)
    filters:
      - variable: DATA_TYPE
        detection: datatype
      - variable: CLIENT
        values:
          - client
          - elasticsearch
          - es
```

## Checklist


pkg/commands/process/settings/rules/javascript/third_parties/elasticsearch.yml
```diff
@@ -0,0 +1,34 @@
...
+  - pattern: |
+      $<CLIENT>.index($<...>$<DATA_TYPE>$<...>)
+    filters:
+      - variable: DATA_TYPE
+        detection: datatype
+      - variable: CLIENT
+        values:
+          - client
+          - elasticsearch
+          - es
```
elsapet, line 11: Do we want to use an auxiliary rule here instead, since I think these values could be anything? 

```javascript
const { Client } = require('@elastic/elasticsearch')
const anything = new Client({
  cloud: { id: '<cloud-id>' },
  auth: { apiKey: 'base64EncodedKey' }
})
```

vjerci: I don't think we support imports atm, so it's would be chaning `new Client`constructor, those intialization properties also don't look very unique to me. 
Honestly I had the same worries but I've figured `index` is unique enough method name. Only thing that came to my mind is index (as in SQL db index), but those shouldn't contain datatypes.

elsapet (approved):

PR #582, opened 2023-02-15
feat(js rules): add segment rule

elsapet:
## Description
Add JavsScript rule for Segment analytics library

Note: does not support segment loaded as a snippet between `<script>` tags in an HTML file e.g. https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/quickstart/

Relates to #504 

## Checklist


pkg/commands/process/settings/rules/javascript/third_parties/segment.yml
```diff
@@ -0,0 +1,64 @@
...
+          - alias
+          - identify
+          - group
+          - page
+          - track
+auxiliary:
+  - id: javascript_third_parties_segment_module
+    patterns:
+      - |
+        require('@segment/analytics-node')
```
vjerci: in javascript all of these are valid, I would try to avoid require as it is a bit outdated way to import
```js
const segment = require('segment')
import segment from 'segment'
// this one is a bit unique as it takes only a part of exported object
import { createClient } from 'segment'
```

vjerci (approved):

pkg/commands/process/settings/rules/javascript/third_parties/segment.yml
elsapet: Agreed; I'll update it 🙏

PR #583, opened 2023-02-15
feat: ruby CWE-502 deserialization of user input

didroe:
## Description

Adds a Ruby rule for deserialization of user input using unsafe methods (ie. methods that can cause code execution). Part of CWE-502.

Rule severity and description TBD

## Checklist


pkg/commands/process/settings/rules/ruby/lang/deserialization_of_user_input.yml
```diff
@@ -0,0 +1,53 @@
...
+      - request
+      - cookies
+      - | # AWS lambda
+        def $<_>($<!>event:, context:)
+        end
+languages:
+  - ruby
+trigger: presence
+severity:
+  default: high # FIXME
```
elsapet, line 38: I'm curious, what is the FIXME for here?

elsapet (approved):

pkg/commands/process/settings/rules/ruby/lang/deserialization_of_user_input.yml
didroe: I don't know what the severity should be

PR #587, opened 2023-02-15
feat: ruby rule for CWE-94 eval using user input

didroe:
## Description

Adds a Ruby rule for CWE-94 when evaluating code using user input.

Rule severity and description TBD

## Checklist


pkg/commands/process/settings/rules/ruby/lang/eval_using_user_input.yml
```diff
@@ -0,0 +1,57 @@
...
+
+    TODO
+
+    ## Remediations
+    TODO
+    <!--
+    ## Resources
+    Coming soon.
+    -->
+  dsr_id: DSR-? # FIXME
```
gotbadger, line 56: field is optional so dont specify if it dosen't have one

didroe: are we dropping DSRs once we've added CWEs? I think it probably has an appropriate DSR but I just haven't added it yet.

elsapet (approved):

Issue #588, opened 2023-02-15
Improve rule testing


gotbadger:
Currently we run e2e for each rule for both dataflow and summary outputs. 

This is becoming increasingly slow as we add more rules and we've also added more test coverage for the language engine so we can be more confident about how most rules will behave.

Proposal:
- [x] profile current test setup and see where time is spent
  - This confimed what we suspected from https://github.com/Bearer/curio/issues/594
- [x] remove dataflow tests https://github.com/Bearer/curio/pull/589
- [x] reduce the number of full e2e tests to cover a handful of cases we think are most complex. https://github.com/Bearer/bearer/pull/621
- [x] introduce integration tests for each rule for the purpose of verification at detector level. https://github.com/Bearer/curio/pull/603

PR #589, opened 2023-02-15
chore: remove dataflow tests

vjerci:
## Description
This pr removes dataflow rules e2e tests as it is redundant test since we have summary tests.

## Related
https://github.com/Bearer/curio/issues/588

## Checklist


integration/rules/javascript_test.go
```diff
@@ -7,56 +7,26 @@ func TestJavascriptLangLoggerSummary(t *testing.T) {
...
 }
 
-func TestJavascriptLangLoggerDataflow(t *testing.T) {
-	t.Parallel()
-	runRulesTest("javascript/lang/logger", "dataflow", "javascript_lang_logger", t)
-}
-
 func TestJavascriptLangSessionSummary(t *testing.T) {
 	t.Parallel()
 	runRulesTest("javascript/lang/session", "summary", "javascript_session", t)
```
didroe, line 12: Can we remove the `"summary"` argument now?

didroe (approved):

integration/rules/javascript_test.go
vjerci: yeah definitely, was hoping to remove rule name too, 
however idea was to do it in another pass once tests are refractored.

elsapet (approved):

PR #592, opened 2023-02-15
docs: add CWE ids to rule metadata

### LINKED ISSUE #591: Add CWE to current rules
To make cross-referencing rules with known risk databases more useful, we're including the CWE IDs to match each rule as `cwe_id` in the metadata. This will gradually phase out the DSR ID.

It's worth noting that some rules match multiple CWEs, and not all CWEs map to OWASP rules.

- [ ] add CWE ids to Ruby rules
- [ ] add CWE ids to JS rules

markmichon:
## Description
Adds CWE ids to each rule. As some contain more than one, an array is used for the cwe_id key. 

## Related
- close #591 

## Checklist


pkg/commands/process/settings/rules/ruby/lang/file_generation.yml
```diff
@@ -71,4 +71,7 @@ metadata:
     Coming soon.
     -->
   dsr_id: DSR-4
+  cwe_id:
+    - 532
+    - 312
```
elsapet: Would this also cover CWE 313 "Cleartext Storage in a File or on Disk"?

pkg/commands/process/settings/rules/ruby/rails/default_encryption.yml
```diff
@@ -32,4 +32,6 @@ metadata:
     Coming soon.
     -->
   dsr_id: DSR-10
+  cwe_id:
+    - 313
```
elsapet: Is this also CWE 311 "Missing Encryption of Sensitive Data"?

Unrelatedly, to confirm my understanding, are we saying this is CWE 313 ("Cleartext Storage in a File or on Disk") rather than CWE 312 ("Cleartext Storage of Sensitive Information") because this case is strictly a SQL database thing?

integration/flags/.snapshots/TestReportFlagsShouldFail-invalid-context-flag
```diff
@@ -37,4 +37,5 @@ General Flags
 
 
 flag error: scan flag error: invalid context argument; supported values: health
+exit status 1
```
elsapet: Ah, you'll probably need to manually remove these `exit status 1` lines from the snapshots to get the integration tests passing here. 

It seems they're generated on local test runs but not by the CI - I'm not sure why but if anyone has an idea, please share 🙏

elsapet (approved):
Left some comments but overall looks good to me 👍

integration/flags/.snapshots/TestReportFlagsShouldFail-invalid-context-flag
cfabianski: I think if you run the integration tests with the binary (like what scripts/run_test does) it will do the right thing. If you do it with the `go run` command, it won’t

elsapet: Ahh the mystery is solved 🤩 Thanks @cfabianski I will try that on my next PR

pkg/commands/process/settings/rules/javascript/lang/session.yml
```diff
@@ -42,4 +42,6 @@ metadata:
     Coming soon.
     -->
   dsr_id: "DSR-5"
+  cwe_id:
+    - 315
```
didroe: It's not stored in a cookie. Maybe the more general parent [CWE-312](https://cwe.mitre.org/data/definitions/312.html) is more appropriate?

pkg/commands/process/settings/rules/javascript/express/insecure_cookie.yml
```diff
@@ -34,5 +34,7 @@ metadata:
 
     ## Resources
     - [Express Security Best Practices](https://expressjs.com/en/advanced/best-practice-security.html#use-cookies-securely)
-  dsr_id: "DSR-5"
+  dsr_id: "DSR-3"
+  cwe_id:
+    - 1004
```
didroe, line 39: We may need to split this rule as I think it also does [CWE-614](https://cwe.mitre.org/data/definitions/614.html)

pkg/commands/process/settings/rules/javascript/lang/exception.yml
```diff
@@ -58,4 +58,6 @@ metadata:
     Coming soon.
     -->
   dsr_id: DSR-5
+  cwe_id:
+    - 209
```
didroe: I think [CWE-210](https://cwe.mitre.org/data/definitions/210.html) (child of 209) is more specific here.

(same comment applies to Ruby)

pkg/commands/process/settings/rules/javascript/lang/file_generation.yml
```diff
@@ -33,4 +33,7 @@ metadata:
     Coming soon.
     -->
   dsr_id: DSR-4
+  cwe_id:
+    - 532
+    - 312
```
didroe: Could use [CWE-313](https://cwe.mitre.org/data/definitions/313.html) (child of 312)

(both comments apply to Ruby also)

pkg/commands/process/settings/rules/javascript/lang/file_generation.yml
```diff
@@ -33,4 +33,7 @@ metadata:
     Coming soon.
     -->
   dsr_id: DSR-4
+  cwe_id:
+    - 532
```
didroe: I don't know if we should include this one. It may be a log, or may not be.

pkg/commands/process/settings/rules/javascript/lang/jwt.yml
```diff
@@ -43,4 +43,6 @@ metadata:
     Coming soon.
     -->
   dsr_id: "DSR-5"
+  cwe_id:
+    - 315
```
didroe: JWT may not be stored in a cookie. Perhaps the more general [CWE-312](https://cwe.mitre.org/data/definitions/312.html) is better here

(same comment applies to Ruby)

pkg/commands/process/settings/rules/javascript/lang/logger.yml
```diff
@@ -95,4 +95,7 @@ metadata:
     Coming soon.
     -->
   dsr_id: "DSR-5"
+  cwe_id:
+    - 209
```
didroe: As we match more than just error logging, maybe [CWE-1295](https://cwe.mitre.org/data/definitions/1295.html) is more appropriate?

(same comment applies to Ruby)

pkg/commands/process/settings/rules/javascript/lang/weak_encryption.yml
```diff
@@ -53,4 +53,7 @@ metadata:
     ## Resources
```
didroe, line 53: While you're in here: I just spotted a typo in the "Remediations" section above, "therefor".

pkg/commands/process/settings/rules/javascript/lang/weak_encryption.yml
```diff
@@ -53,4 +53,7 @@ metadata:
...
     - [NodeJS Crypto Module](https://nodejs.org/api/crypto.html#cryptocreatehmacalgorithm-key-options)
   dsr_id: "DSR-5"
+  cwe_id:
+    - 331
```
didroe: I don't think this one necessarily applies

pkg/commands/process/settings/rules/javascript/lang/weak_encryption.yml
```diff
@@ -53,4 +53,7 @@ metadata:
...
+    - 326
```
didroe: Perhaps [CWE-327](https://cwe.mitre.org/data/definitions/327.html) instead of, or as well as, this one

(both comments also apply to Ruby)

pkg/commands/process/settings/rules/javascript/react/google_analytics.yml
```diff
@@ -28,4 +28,7 @@ metadata:
     Coming soon.
     -->
   dsr_id: DSR-1
+  cwe_id:
+    - 200
```
didroe: I don't think there's any problem in doing this, but as a general point: 201 is a child of 200 (200 is implied by 201) so I'm not sure we need to list both 

(same comment applies to other rules)

pkg/commands/process/settings/rules/ruby/lang/ssl_verification.yml
```diff
@@ -41,4 +41,6 @@ metadata:
     Coming soon.
     -->
   dsr_id: DSR-2
+  cwe_id:
+    - 319
```
didroe: It's not cleartext, just not fully verified. I think [CWE-295](https://cwe.mitre.org/data/definitions/295.html) is more appropriate

pkg/commands/process/settings/rules/ruby/lang/weak_encryption_with_data.yml
```diff
@@ -148,4 +148,8 @@ metadata:
     ## Resources
     - [BCrypt Explained](https://dev.to/sylviapap/bcrypt-explained-4k5c)
   dsr_id: DSR-7
+  cwe_id:
+    - 311
```
didroe: I wouldn't say we're missing encryption. I think 326 covers the sensitive data part anyway so maybe drop this?

pkg/commands/process/settings/rules/ruby/rails/default_encryption.yml
didroe: I think 312 is more general here. It might not be a DB stored on disk. Also 312 is a child of 311 so I'd say that's implied without adding it explicitly.

pkg/commands/process/settings/rules/ruby/third_parties/bigquery.yml
```diff
@@ -3,15 +3,15 @@ patterns:
       $<INSERTER>.insert($<...>$<DATA_TYPE>$<...>)$<...>
     filters:
       - either:
-        - variable: INSERTER
-          detection: ruby_third_parties_bigquery_dataset
-        - variable: INSERTER
-          detection: ruby_third_parties_bigquery_insert_async
-        - variable: INSERTER
-          detection: ruby_third_parties_bigquery_table
+          - variable: INSERTER
```
didroe, line 6: Unwanted change here?

new/detector/implementation/ruby/.snapshots/TestRubyObjectDetector-object_assignment
```diff
@@ -7,13 +7,4 @@
           matchnode: {}
           data:
             name: first_name
-- position: "1:8"
```
didroe, line 10: This seems unexpected (also the `object_parent_pair` one)

pkg/commands/process/settings/rules/ruby/rails/default_encryption.yml
markmichon: Ah, I missed the `stored_data_types` trigger on this (we don't document it) so I was avoiding any CWE about sensitive data. Will update to 312.

pkg/commands/process/settings/rules/ruby/third_parties/bigquery.yml
markmichon: hmm. prettier may have fixed it. Only did it to this block and this file, so assuming there was some inconsistency. We might want to set up an editorconfig or equivalent at some point for the repo.

PR #599, opened 2023-02-16
feat: log when loading rules

### LINKED ISSUE #594: Slow startup time
## Description & Reproduction

There is a 5-10 second delay when starting up. 

## Expected Behavior

There should be no delay, or a log should be produced so it's clear something is happening.

## Actual Behavior

There is a delay with no output.

## Possible Fix

The delay could be caused by compiling all the rule patterns. Tree sitter queries seem quite slow to compile.

### to do:

- [x] Split each rule initialization in a separate goruting in order to better utilise cpu (2x increase) https://github.com/Bearer/bearer/pull/618

didroe:
## Description

Adds a log message to let the user know rules are being loaded/compiled. 

- Closes https://github.com/Bearer/bearer/issues/594

## Checklist

### DIFF
--- a/pkg/commands/artifact/run.go
+++ b/pkg/commands/artifact/run.go
@@ -23,6 +23,7 @@ import (
 	"github.com/bearer/bearer/pkg/commands/process/worker/work"
 	"github.com/bearer/bearer/pkg/flag"
 	reportoutput "github.com/bearer/bearer/pkg/report/output"
+	"github.com/bearer/bearer/pkg/util/output"
 	outputhandler "github.com/bearer/bearer/pkg/util/output"
 
 	"github.com/bearer/bearer/pkg/types"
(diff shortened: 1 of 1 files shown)


new/detector/composition/composition.go
```diff
@@ -1,20 +1,80 @@
...
 	reportdetections "github.com/bearer/curio/pkg/report/detections"
 	"github.com/bearer/curio/pkg/report/detectors"
 	"github.com/bearer/curio/pkg/report/schema"
 	"github.com/bearer/curio/pkg/report/source"
 	"github.com/bearer/curio/pkg/util/file"
+	"github.com/bearer/curio/pkg/util/output"
 )
 
+func CompileRules(config *settings.Config) (map[string][]detectortypes.Detector, error) {
+	output.StdErrLogger().Msg("Compiling rules")
```
gotbadger: Need to respect Quiet flag here I think?

gotbadger (approved):

PR #603, opened 2023-02-17
chore: redo rules tests

vjerci:
## Description
This Pr optimises rules tests in a way that we no longer perform e2e test but do integration tests for them instead.

Trough profiling we've found that biggest chunk cpu time is spent on compiling all the rules.
I've optimised it in a way that we use worker to compile all the rules only once, then reuse that worker.
It isn't without it's downsides. Because we reuse single testing worker, and because we use viper for default config, we can no longer run tests in parallel.
Getting tests to run in parallel is limited to single test case which translates to single rule (it usually has multiple files which run in parallel)
- We might in future do some extra work to have a worker per language or some other logical split to further improve performance 

However I've manged to improve the total time it takes to run our rules test by 2x-3x and allow us to run tests with debugger

---

### summary changes
There are cases where we detect 2 different rules in summary.
Eg. of those rules is `ruby_weak_encryption` and `ruby_weak_encryption_with_data` 

---

### stats

localmachine (macbook pro)
old_summary - 187.077s
new - 51.512s

github runner
old_summary - 332s
new - 183.678s

---

### related

https://github.com/Bearer/curio/issues/588

## Checklist


integration/rules/javascript_test.go
```diff
@@ -2,122 +2,96 @@ package integration_test
 
 import "testing"
 
+const javascriptRulesPath string = "../../pkg/commands/process/settings/rules/javascript/"
+
 func TestJavascriptLangLoggerSummary(t *testing.T) {
```
elsapet: Suggestion to remove `Summary` from the naming to avoid possible confusion

```suggestion
func TestJavascriptLangLogger(t *testing.T) {
```

elsapet (approved):
🙏

integration/rules/javascript_test.go
vjerci: done, nice suggestion 👍

PR #618, opened 2023-02-20
feat: optimise compilation time

vjerci:
## Description
This pr offloads each detector creation into a separate gorutine, allowing more efficient cpu utilisation on multi cpu machines.
When benchmarking performance on macbook pro a single js rule test took 7.8s to execute without this fix and 3.4s with this fix. 
3.4s startup time is still a lot but it is also a 2x performance improvement compared to what we previously had

## Checklist

### DIFF
--- a/new/detector/composition/javascript/javascript.go
+++ b/new/detector/composition/javascript/javascript.go
@@ -75,7 +75,7 @@ func New(rules map[string]*settings.Rule, classifier *classification.Classifier)
 	}
 
 	detectorsLen := len(jsRules) + len(staticDetectors)
-	reciever := make(chan types.DetectorInitResult, detectorsLen)
+	receiver := make(chan types.DetectorInitResult, detectorsLen)
 
 	var detectors []detectortypes.Detector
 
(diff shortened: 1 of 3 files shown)


new/detector/composition/javascript/javascript.go
```diff
@@ -64,16 +65,30 @@ func New(rules map[string]*settings.Rule, classifier *classification.Classifier)
...
+	jsRules := map[string]*settings.Rule{}
+	for ruleName, rule := range rules {
+		if !slices.Contains(rule.Languages, "javascript") {
+			continue
+		}
+		jsRules[ruleName] = rule
+	}
+
+	detectorsLen := len(jsRules) + len(staticDetectors)
+	reciever := make(chan types.DetectorInitResult, detectorsLen)
```
elsapet:
```suggestion
	receiver := make(chan types.DetectorInitResult, detectorsLen)
```

elsapet (approved):
Tiny typo but looks good

new/detector/composition/javascript/javascript.go
vjerci: thank you

PR #622, opened 2023-02-20
feat(JS rules): express sendFile with request data

elsapet:
## Description

Raise a warning when a filename or path is determined by user input (in express, request data) when sending a file. 

Relates to #606 

## Checklist


pkg/commands/process/settings/rules/javascript/express/external_file_upload.yml
```diff
@@ -0,0 +1,21 @@
+patterns:
+  - pattern: res.sendFile($<EXPRESS_REQ>)
```
elsapet, line 2: This also matches on something like the following (i.e. further arguments following `$<EXPRESS_REQ>`), which is unexpected to me since there is no `$<...>`. 

Or maybe I am misunderstanding how the `$<_>` works in the `EXPRESS_REQ` pattern.

Any tips on how to exclude this case, if possible? Ping @didroe or @vjerci 🙏 

```javascript
res.sendFile(req.params.file, { root: path.join(__dirname, 'public') })
```

Edit: here are the generated patterns, for reference: 

```
{"level":"debug","time":"2023-02-21T13:58:22+02:00","message":"compiled pattern req.$<_> -> ([(member_expression . [ (identifier )] @param1  . (_) .)] @root)"}
```

```
{"level":"debug","time":"2023-02-21T13:58:22+02:00","message":"compiled pattern res.sendFile($<EXPRESS_REQ>) -> ([(call_expression . [(member_expression . [ (identifier )] @param1  . [ (property_identifier )] @param2 .)] . [(arguments  . [(_)] @param3 )] .)] @root)"}
```

vjerci: not sure if that's the expected behaviour either

pkg/commands/process/settings/rules/javascript/express/external_file_upload.yml
```diff
@@ -0,0 +1,21 @@
...
+  - javascript
+auxiliary:
+  - id: javascript_express_request_obj
+    patterns:
+      - req.$<_>
+trigger: presence
+severity:
+  default: "warning"
+metadata:
+  description: "Do not allow user input to control filename or path."
```
didroe: We should now follow the convention from https://github.com/Bearer/bearer/pull/627

pkg/commands/process/settings/rules/javascript/express/external_file_upload.yml
didroe: I would expect the pattern to only match a single argument. Let's have a look at this one together

cfabianski (approved):

PR #632, opened 2023-02-21
feat: insecure http password and weak encryption

vjerci:
## Description
This pr adds insecure http password rule for rails
And enriches weak encryption rule for ruby

## Checklist


integration/rules/ruby_test.go
```diff
@@ -70,6 +70,10 @@ func TestRubyRailsInsecureCommunication(t *testing.T) {
 	getRunner(t).runTest(t, rubyRulesPath+"rails/insecure_communication")
 }
 
+func TestRubyRailsInsecureHTTPPassowrd(t *testing.T) {
```
elsapet, line 99:
```suggestion
func TestRubyRailsInsecureHTTPPassword(t *testing.T) {
```

pkg/commands/process/settings/rules/ruby/lang/weak_encryption.yml
```diff
@@ -34,6 +34,11 @@ patterns:
           - export
           - to_pem
           - to_s
+  - pattern: |
+      OpenSSL::PKey::RSA.new($<LENGTH>)
```
elsapet: Do we want DSA here too? 

https://ruby-doc.org/stdlib-3.1.1/libdoc/openssl/rdoc/OpenSSL/PKey/DSA.html

pkg/commands/process/settings/rules/ruby/lang/weak_encryption/testdata/openssl_dsa.rb
```diff
@@ -1,5 +1,11 @@
+size = 2048
 cipher = OpenSSL::Cipher.new('aes-128-cbc')
-dsa_encrypt = OpenSSL::PKey::DSA.new(2048)
+rsa_encrypt = OpenSSL::PKey::RSA.new(size)
```
elsapet: We'd want these changes in the RSA file: https://github.com/Bearer/bearer/blob/main/pkg/commands/process/settings/rules/ruby/lang/weak_encryption/testdata/openssl_rsa.rb

elsapet (commented):
I think we can support both RSA and DSA libraries here

pkg/commands/process/settings/rules/ruby/lang/weak_encryption.yml
vjerci: Had to do some googling to figure it out,  
found some resource that said DSA 1024 are secure enough.

pkg/commands/process/settings/rules/ruby/lang/weak_encryption/testdata/openssl_dsa.rb
vjerci: nice catch

elsapet (approved):

PR #643, opened 2023-02-22
feat(JS rules): add rule for express cross-site scripting

elsapet:
## Description
Add rule for XSS vulnerabilities in Express JS
- Sequelize queries containing req data
- res.write and res.send containing req data

## Checklist


pkg/commands/process/settings/rules/javascript/express/cross_site_scripting.yml
```diff
@@ -0,0 +1,38 @@
...
+      res.write($<EXPRESS_REQ>)
+    filters:
+      - variable: EXPRESS_REQ
+        detection: javascript_express_cross_site_scripting_request_obj
+  - pattern: |
+      res.send($<EXPRESS_REQ>)
+    filters:
+      - variable: EXPRESS_REQ
+        detection: javascript_express_cross_site_scripting_request_obj
+  - pattern: |
```
didroe: I don't think this one should be XSS as it's not being put into the response. I would make this part of a rule covering [CWE-89](https://cwe.mitre.org/data/definitions/89.html)

elsapet: Agreed; thanks for the catch!
CWE-89 PR here: https://github.com/Bearer/bearer/pull/648

didroe (approved):

PR #645, opened 2023-02-22
feat(JS rules): express.js path traversal

elsapet:
## Description
Add rule against path traversal vulnerability - to catch whenever request data is included in path.join or path.resolve

Relates to #606 

## Checklist


pkg/commands/process/settings/rules/javascript/express/path_traversal.yml
```diff
@@ -0,0 +1,27 @@
...
+trigger: presence
+severity:
+  default: "warning"
+metadata:
+  description: "Possible path traversal vulnerability detected."
+  remediation_message: |
+    ## Description
+    TODO
+  cwe_id:
+    - 611
```
didroe: This seems more appropriate:
```suggestion
    - 22
```

didroe (approved):

PR #649, opened 2023-02-23
feat: add rails open redirect rule

vjerci:
## Description
Add rails open redirect rule

Patterns:
```
  - pattern: |
      redirect_to(params[$<_>])
```
## Checklist


elsapet (approved):

pkg/commands/process/settings/rules/ruby/rails/open_redirect.yml
```diff
@@ -0,0 +1,16 @@
+patterns:
+  - pattern: |
+      redirect_to(params[$<_>])
```
didroe: It would be good to use a similar definition of "user input" to the other rules:

https://github.com/Bearer/bearer/blob/main/pkg/commands/process/settings/rules/ruby/lang/ftp_using_user_input.yml#L20-L27

(AWS lambda case probably isn't relevant though)

vjerci: nice 👍

PR #650, opened 2023-02-23
feat: jwt hardcoded secret

vjerci:
## Description

Update jwt cwe

## Checklist


elsapet (approved):

pkg/commands/process/settings/rules/javascript/lang/jwt.yml
```diff
@@ -45,4 +45,5 @@ metadata:
   dsr_id: "DSR-5"
   cwe_id:
     - 312
+    - 798
```
didroe: I'm not sure about this. Sensitive data isn't necessarily credentials.

vjerci: yeah sorry, I thought it was the same rule, it seems it's not.
It is fixed now with new rule.

elsapet (approved):

PR #652, opened 2023-02-23
feat: add jwt weak encryption

vjerci:
## Description
This pr adds support for js jwt weak encryption detection 

## Checklist


pkg/commands/process/settings/rules/javascript/lang/jwt_weak_encryption.yml
```diff
@@ -0,0 +1,32 @@
...
+      jwt.sign($<_>, $<_>, {
+        algorithm: [$<...>"none"$<...>]
+      })
+languages:
+  - javascript
+trigger: presence
+severity:
+  default: critical
+skip_data_types:
+  - "Unique Identifier"
```
elsapet: I don't think this is needed since trigger is `presence`
```suggestion
```

elsapet (approved):

PR #664, opened 2023-02-24
feat(JS rules): add AWS dynamodb query injection

elsapet:
## Description
Add rule to catch general (not SQL-lang specific) query injection vulnerability, namely AWS DynamoDB queries in AWS Lambda

CWE-943

Relates to #606 

## Checklist


pkg/commands/process/settings/rules/javascript/aws_lambda/query_injection/testdata/ok_no_injection.js
```diff
@@ -0,0 +1,18 @@
...
+exports.handler = async function(event, _context) {
+  const params = {
+    KeyConditionExpression: "Title = " + getTitle(),
+    FilterExpression: "contains (Author, :name)",
+    ExpressionAttributeValues: {
+      ":name": { S: getAuthorName() },
+    },
+    ProjectionExpression: "Title, Author",
+    TableName: "BOOKS_TABLE",
+  };
```
elsapet, line 15: It's safe to reference the event object in this kind of hash but I couldn't get the pattern to work for it. Anyone any ideas 😅

elsapet: ^ After discussion with @cfabianski we're fine to fail on these cases, as it is still user-defined input coming into the query

cfabianski (approved):

PR #673, opened 2023-02-27
fix: fix false positives on req detection

elsapet:
## Description

Stricter pattern match for detecting problematic request objects. Combined with fix here https://github.com/Bearer/bearer/pull/622/commits/0c1ef456b20e20d2f4b168e6f981a7c1a086b67a this should address a lot of the false positives seen in battle testing. 

Relates to #606 

## Checklist


pkg/commands/process/settings/rules/javascript/express/cross_site_scripting/testdata/no_xss.js
```diff
@@ -1,6 +1,24 @@
...
+})
+
+app.get("/good-3", () => {
+  const results = newResult(search);
+
+  res.send(results);
+})
+
+app.get("/good-4", () => {
+  return res.send({ success: false, text: `User ${req.params.user_id} not found` });
```
cfabianski: I guess it could also be `req.params["user_id"]` which wouldn't be found as it it. In which conditions did you see it fail before / raise false positives?

elsapet: True, it won't catch `req.params["user_id"]`. Will add some test cases for that one. 

This fix is more for the false positives that have come out of the batte testing: https://www.notion.so/bearer/Curio-battle-test-review-0-25-Javascript-55dcf15da8ff49088218ee852f7498bf

cfabianski:
```javascript
res.send({
  body: inspect(req.body, { depth: Infinity, maxStringLength: 100 })
})
```

We were discussing with @didroe about a way to mark a detection as `not-risky`. I guess this is another instance of it.
🤔

cfabianski (approved):

pkg/commands/process/settings/rules/javascript/express/cross_site_scripting/testdata/no_xss.js
elsapet: Resolving so PR can be merged :)

PR #685, opened 2023-02-28
fix(JS rules): exclude async calls when matching on data types for logger and exception leak rules

### LINKED ISSUE #674: False positives with data type detections
## Description & Reproduction

Javascript rules `javascript_lang_exception` and `javascript_lang_logger` trigger if there is a property access in the exception or logger call and the receiver object has data types.

Code sample (pseudo but the structure is represented): 

```javascript
// exception
async function run(err) {
  const resp = await this.getResponse({
    data: {
      name: this.name,
      email: this.email
    }
  })

  if (resp.error != nil) {
    throw new Error(resp.error)
  }

  return data
}
```

```javascript
// logger
async function run(err) {
  const resp = await this.getResponse({
    data: {
      name: this.name,
      email: this.email
    }
  })

  if (resp.error != nil) {
    console.log(resp.error)
  }

  return data
}
```

## Expected Behavior

The rules should not be triggered in these cases since the exception call or logger call do not involve a data type

## Actual Behavior

The rules are triggered

## Further examples

The exception rule is also triggered by the following code (unexpected), which may be a related issue: 

```javascript
const funcAlice = () => {
	return new Promise((resolve, reject) => {})
};

const funcBob = () => {
		let data = {
			email:      'admin@example.com',
			name:       'Administrator',
		};
};

module.exports = function () {
	return funcAlice()
		.then(funcBob)
};
```

And the Ruby Rails session rule (id `ruby_rails_session`) is triggered by the following code, when we wouldn't expect it to trigger: 

```ruby
user = User.find_by_name(params[:username])
session[:user_id] = user.id 
```

## Your Environment

- Output of 'bearer version':
```
bearer version: current `main`
sha: 84fd36b5aa477747e508c2a32c48ab54238ad304
```

elsapet:
## Description

The logger and exception leak rules were triggering false positives on a common JS pattern: 

```js
var result = await someAsyncCall({
  email: currentUser.email,
  name: currentUser.name
})

if (result.error != null) {
  console.log(result.error)
  throw new SomeError(result.error)
}
```

In this PR we update the rule patterns to exclude async function calls when we are matching data types for these two rules only. This means we may miss on some true positives but the majority of triggers of these rules in their current state would be false. 

Closes #674 

## Checklist


vjerci (approved):

cfabianski (changes requested):
I think the real issue is that we get a relationship between results and error

await should have blocked the flow IMO

wdyt @didroe ?

didroe:
@cfabianski I agree. There shouldn't be any link between `result` and the function arguments. So the rule for throw should just be considering an object called `result` with a single `error` property, and not trigger.

I think this is probably an issue for Ruby too

didroe:
@cfabianski actually I'm having second thoughts about this. We treat functions like identity functions at the moment. ie. we assume their return value contains or is influenced by the inputs. In the absence of more complex flow analysis, that's maybe not a bad thing (though prone to false +ves).

In this specific case, none of that should matter anyway as we shouldn't be considering `result`. What I would expect to happen is that there is an object detection for `result.error` with just an `error` property, and when it finds that detection it should stop descending in the tree. So we shouldn't be finding the detection for `result`. We should address that first (which will fix this particular issue) and then we can further discuss return values.

cfabianski:
> @cfabianski actually I'm having second thoughts about this. We treat functions like identity functions at the moment. ie. we assume their return value contains or is influenced by the inputs. In the absence of more complex flow analysis, that's maybe not a bad thing (though prone to false +ves).
> 
> In this specific case, none of that should matter anyway as we shouldn't be considering `result`. What I would expect to happen is that there is an object detection for `result.error` with just an `error` property, and when it finds that detection it should stop descending in the tree. So we shouldn't be finding the detection for `result`. We should address that first (which will fix this particular issue) and then we can further discuss return values.

Yeah that's exactly what I was trying to explain as well but didn't manage to express properly. 👍

elsapet:
Closing this because we will look for a less naïve solution (see comments)

PR #689, opened 2023-02-28
feat(JS rules): Add CWE 916 weak password encryption

elsapet:
## Description
Add new JS weak encryption rule that targets passwords (CWE-916)
Update existing JS weak encryption rule:
* Include CryptoJS module methods
* Don't match on passwords

Relates to #606 

## Checklist


pkg/commands/process/settings/rules/javascript/lang/weak_encryption.yml
```diff
@@ -6,6 +6,29 @@ patterns:
         detection: create_hash
       - variable: DATA_TYPE
         detection: datatype
+  - pattern: |
+      CryptoJS.$<METHOD>($<PASSWORD>)
```
cfabianski: Should we use another name other than PASSWORD here if the idea is to skip passwords anyway?

elsapet: Good spot, thanks. Copy pasta mistake 🍝 🙈

cfabianski (approved):

PR #699, opened 2023-03-01
feat(ruby rules): rule for hardcoded secret

didroe:
## Description

Adds a Ruby rule for detecting hardcoded secrets.

To support this, we introduce a new built-in detector for string literals. Existing rules are updated to use this too - the previous way they were matching is to try to look for a typed variable but (a) this doesn't work if using variables/constants and (b) we'd like to remove the typing of variables at some point.

## Checklist


cfabianski (approved):
Really good! Thanks

PR #701, opened 2023-03-01
fix: don't report nested datatype detections

didroe:
## Description

We don't keep descending down the tree for object detections once one is found, but we currently do for datatype detections (which are just classified objects).

This change makes the behaviour consistent and prevents us finding datatypes from a parent object when . ie.

```javascript
let user = { email: "", other: "" }

user.other // this expression would include "email" prior to this change and will just be "other" now
```

## Checklist


didroe:
Closing until we discuss and figure out the best way to deal with this

PR #707, opened 2023-03-02
feat(JS ruels): add insecure CORS rule for express

elsapet:
## Description
* Add expressJS rule to catch when origin is set by user input

Relates to #606 

## Checklist


pkg/commands/process/settings/rules/javascript/express/insecure_allow_origin.yml
```diff
@@ -0,0 +1,62 @@
...
+    patterns:
+      - pattern: |
+          { $<ORIGIN>: $<USER_INPUT> }
+        filters:
+          - variable: ORIGIN
+            values:
+              - '"Access-Control-Allow-Origin"'
+              - '"access-control-allow-origin"'
+              - "'Access-Control-Allow-Origin'"
+              - "'access-control-allow-origin'"
```
cfabianski: We could use a regex here, might simplify that a bit
see https://github.com/Bearer/bearer/pull/575

cfabianski (approved):

PR #712, opened 2023-03-03
feat(JS rules): add template render rule patterns for express js

elsapet:
## Description
* Add rule to catch insecure template rendering with custom template modules

Note: some of these modules can be set as the default (e.g. `res.render`) and these cases we expect to be caught by the existing `javascript_express_external_resource` rule.

Relates to #606

## Checklist


cfabianski (approved):
Looking good

PR #715, opened 2023-03-03
feat(JS rules): add expressjs rule for missing HTTPS protocol

elsapet:
## Description
Add rule to trigger in case an app is using `http.createServer()` and not `https.createServer()`

## Checklist


cfabianski (approved):

pkg/commands/process/settings/rules/javascript/express/https_protocol_missing.yml
```diff
@@ -0,0 +1,32 @@
...
+  - javascript
+trigger: absence
+severity:
+  default: "medium"
+metadata:
+  description: "Security misconfiguration detected."
+  remediation_message: |
+    ## Description
+
+    Helmet can help protect your app from some well-known web vulnerabilities by setting HTTP headers appropriately.
```
gotbadger: Description seems wrong here?

elsapet: 🙈

elsapet: Thanks for the catch @gotbadger - updated 👍

cfabianski: Nice catch @gotbadger

PR #723, opened 2023-03-03
feat(ruby rules): add rails rule for permissive regex validation

didroe:
## Description

Adds a Rails rule for CWE-625, finding permissive regular expressions in format validations.

## Checklist


cfabianski (approved):
Very well done

PR #743, opened 2023-03-06
fix(JS rules): fix insert HTML rule

elsapet:
## Description
* Update `not` filter that uses `string_literal` detector, following bug fix to `string_literal` detector
* Exclude `React` as source for certain methods e.g. `createElement`

## Checklist

### DIFF
--- a/pkg/commands/process/settings/rules/javascript/lang/dangerous_insert_html.yml
+++ b/pkg/commands/process/settings/rules/javascript/lang/dangerous_insert_html.yml
@@ -20,8 +20,12 @@ patterns:
           variable: DATA
           detection: javascript_dangerous_insert_html_sanitzed_input
   - pattern: |
-      $<_>.$<METHOD>($<...>$<DATA>$<...>)
+      $<SOURCE>.$<METHOD>($<...>$<DATA>$<...>)
     filters:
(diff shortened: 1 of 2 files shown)


pkg/commands/process/settings/rules/javascript/lang/dangerous_insert_html.yml
```diff
@@ -41,6 +45,11 @@ auxiliary:
           - variable: STRING_LITERAL
             detection: string_literal
             contains: false
+      - pattern: |
+          $<EMPTY_STRING>
+        filters:
+          - variable: EMPTY_STRING
+            regex: (""|'')
```
vjerci:
```suggestion
            regex: (""|''|``)
```

vjerci (approved):

pkg/commands/process/settings/rules/javascript/lang/dangerous_insert_html.yml
elsapet: Great catch; thanks 😎

cfabianski:
Won't be necessary as we are going to disable that rule https://github.com/Bearer/bearer/pull/745

PR #754, opened 2023-03-08
fix(JS Express rules): improve default cookie rule

elsapet:
## Description
* Only trigger default cookie rule if there's a hash present

## Checklist

### DIFF
--- a/new/language/implementation/javascript/javascript.go
+++ b/new/language/implementation/javascript/javascript.go
@@ -239,7 +239,7 @@ func (implementation *javascriptImplementation) IsRootOfRuleQuery(node *tree.Nod
 }
 
 func (implementation *javascriptImplementation) PatternNodeTypes(node *tree.Node) []string {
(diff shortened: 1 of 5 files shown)


new/language/implementation/javascript/javascript.go
```diff
@@ -239,6 +239,10 @@ func (implementation *javascriptImplementation) IsRootOfRuleQuery(node *tree.Nod
 }
 
 func (implementation *javascriptImplementation) PatternNodeTypes(node *tree.Node) []string {
+	if node.Type() == "statement_block" && node.Parent().Type() == "program" {
```
didroe: I think we should also check for children, that will allow us to still match against statement blocks in patterns:

```suggestion
	if node.Type() == "statement_block" && node.Parent().Type() == "program" && node.NamedChildCount() == 0 {
```

didroe (approved):

PR #759, opened 2023-03-09
fix(JS Express rules): tighten path traversal rule

elsapet:
## Description
A "fix" for path traversal is to replace dangerous characters like `.../.../` with empty strings.
We tighten the JS Express rule around path traversal to exclude cases where we are calling replace on the user input.

## Checklist

### DIFF
--- a/pkg/commands/process/settings/rules/javascript/express/path_traversal.yml
+++ b/pkg/commands/process/settings/rules/javascript/express/path_traversal.yml
@@ -8,6 +8,9 @@ patterns:
           - resolve
       - variable: USER_INPUT
         detection: javascript_express_path_traversal_request_obj
+      - not:
+          variable: USER_INPUT
+          detection: javascript_express_path_traversal_sanitized_user_input
 auxiliary:
   - id: javascript_express_path_traversal_request_obj
     patterns:
@@ -16,6 +19,10 @@ auxiliary:
       - req.body
       - req.cookies
       - req.headers
+  - id: javascript_express_path_traversal_sanitized_user_input
+    patterns:
+      - $<_>.replace($<_>, "")
+      - $<_>.replace($<_>, '')
 languages:
   - javascript
 trigger: presence
--- a/pkg/commands/process/settings/rules/javascript/express/path_traversal.yml
+++ b/pkg/commands/process/settings/rules/javascript/express/path_traversal.yml
@@ -32,12 +32,36 @@ metadata:
   remediation_message: |
     ## Description
     Allowing unsanitized user input in path resolution methods means an attacker could gain access to files and folders outside of the intended scope.
-    <!--
+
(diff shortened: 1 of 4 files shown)


pkg/commands/process/settings/rules/javascript/express/path_traversal.yml
```diff
@@ -8,6 +8,9 @@ patterns:
           - resolve
```
soheil, line 8: Please add a comment for the types of `req` that are considered for User Input in the `javascript_express_path_traversal_request_obj` pattern.

pkg/commands/process/settings/rules/javascript/express/path_traversal.yml
```diff
@@ -8,6 +8,9 @@
...
       - variable: USER_INPUT
         detection: javascript_express_path_traversal_request_obj
+      - not:
+          variable: USER_INPUT
+          detection: javascript_express_path_traversal_sanitized_user_input
 auxiliary:
   - id: javascript_express_path_traversal_request_obj
     patterns:
```
soheil, line 16: Please clarify the expected safe characters for `replace` calls in the `javascript_express_path_traversal_sanitized_user_input` pattern.

soheil (commented):
Overall, this seems like a good change that should tighten the path traversal rule.

A few questions and comments based on the diff provided:

In `javascript_express_path_traversal_request_obj` pattern, which types of `req` are being considered for User Input? May be it is better to add a descriptive comment here?

In the added pattern `javascript_express_path_traversal_sanitized_user_input`, is it intended to restrict the `USER_INPUT` object out of `replace` calls? It would be helpful if `replace` parameters be added here in order to have a clear understanding of the expected safe characters.

Assuming that the related tests have all passed successfully, this change seems reasonable and helpful.

pkg/commands/process/settings/rules/javascript/express/path_traversal.yml
elsapet: Thanks - good idea. I've updated the `remediation_message` to provide an example of the expected safe characters here https://github.com/Bearer/bearer/pull/759/commits/8b7e363741620ad681db772a922dc46df6d2cda2

pkg/commands/process/settings/rules/javascript/express/path_traversal.yml
elsapet: The `req` types that we consider User Input are the following: 
https://github.com/Bearer/bearer/blob/d6e2dbe928007211e8f02b9b376b9457d63327c6/pkg/commands/process/settings/rules/javascript/express/path_traversal.yml#L12-L18
I feel the patterns are descriptive enough themselves. What sort of comment would you like to see here?

gotbadger (approved):

PR #767, opened 2023-03-10
fix: split integration tests

elsapet:
## Description

Separate github integration test workflow into two, one for ruby tests and one for javascript tests.

Note: this should be considered as a temporary fix only, to address CI timeouts on `main` branch. A more sophisticated testing solution will be considered during Milestone 3 ("Improve rule integration testing workflows") https://github.com/Bearer/bearer/issues/766

## Checklist


.github/workflows/ruby_integration_test.yml
```diff
@@ -0,0 +1,27 @@
+name: Javascript Integration Tests
```
gotbadger:
```suggestion
name: Ruby Integration Tests
```

elsapet: 🙈 🙏

gotbadger (approved):

PR #771, opened 2023-03-10
fix(Ruby rules): tighten ruby hardcoded secret rule pattern

### LINKED ISSUE #760: Improve Ruby hardcoded secret rule
## Description & Reproduction

We're seeing some false positives with this rule, in some cases around the `Passwordless` library methods, which are matching by the current rule patterns

## Expected Behavior

The following code sample should not trigger the hardcoded secret rule:

```ruby
Passwordless.default_from_address = "some_address@bear.com"
```

## Actual Behavior

The rule is triggered by the sample above

## Possible Fix

Tighten the rule pattern given these false positive cases

## Your Environment

- Output of 'bearer version':
```
bearer version: 1.00.0
sha: 77d701df4e71763ec6677aea09fcb1e62e97bf0d
```

elsapet:
## Description
Add an end word boundary to exclude Passwordless case

Closes #760 

## Checklist

### DIFF
--- a/pkg/commands/process/settings/rules/ruby/lang/hardcoded_secret.yml
+++ b/pkg/commands/process/settings/rules/ruby/lang/hardcoded_secret.yml
@@ -3,7 +3,7 @@ patterns:
       $<NAME> = $<STRING_LITERAL>
     filters:
       - variable: NAME
-        regex: (?i)password|api_?key|secret
+        regex: (?i)(password|api_?key|secret)\b
       - variable: STRING_LITERAL
         detection: string_literal
         contains: false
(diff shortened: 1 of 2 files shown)


pkg/commands/process/settings/rules/ruby/lang/hardcoded_secret.yml
```diff
@@ -3,15 +3,15 @@ patterns:
       $<NAME> = $<STRING_LITERAL>
     filters:
       - variable: NAME
-        regex: (?i)password|api_?key|secret
+        regex: (?i)(password|api_?key|secret)\\b
```
elsapet: The below felt too restrictive, but it would bring this case in line with the string variants e.g. `(?i)['"](password|api_?key|secret)['"]`. 
Any thoughts or preferences, @didroe ? 

```suggestion
        regex: \\b(?i)(password|api_?key|secret)\\b
```

cfabianski (approved):
👏

pkg/commands/process/settings/rules/ruby/lang/hardcoded_secret.yml
elsapet: Resolving for now but still open to any comments 🙏

PR #789, opened 2023-03-14
feat: add support for object deconstructing.

vjerci:
## Description
This pr adds support for object deconstructing.
It adds minimal support in a way we only support first level of object deconstructing.

---

### Supported findings

-  `const` will detect both `let` and `const`, they are considered the same
- `var` needs to be specified - it's different than `const` or `let` 

Rule:
```
const { $<!>$<_> } = req.params
```
Matching testcases:
```
const { user } = req.params
let { user } = req.params
```

Rule:
```
var { $<!>$<_> } = req.params
```
Matching testcases:
```
var { user } = req.params
```

---

### Unsuported findings
Nested cases
- we don't support cases like these
```
var { user: { admin, superuser } } = req.params
```

---

## Related
https://github.com/Bearer/bearer/issues/698

## Checklist


new/detector/composition/javascript/testdata/testcases/object-deconstructing/deconstructing.js
```diff
@@ -0,0 +1,2 @@
+const { user } = req.params;
```
cfabianski, line 1: Do we support things like `const { firstName,  lastName } = req.params` as well?
It's not clear to be based on what I can see

vjerci: we do, I've added test case for it

cfabianski: I don't see the test case :/
Could you send me the link to it please?

elsapet (approved):

cfabianski (approved):

cfabianski (approved):
LGTM. Good job

Issue #545, opened 2023-02-09
Improve external rule error messages


markmichon:
Currently, if there is a problem with the pattern for a custom rule you get the following error:
```json
{"level":"fatal","time":"2023-02-09T11:38:33-08:00","message":"Failed to start curio, error with your configuration custom detector error: failed to instantiate composition ruby:failed to create rule custom_rule_name: error compiling pattern: failed to build: error parsing pattern at 1:16: CurioVar1"}
exit status 1
```

Example pattern that triggered the error:
```
patterns:
  - pattern: |
      JWT.$<_>({$<DATA_TYPE>})
    filters:
      - variable: DATA_TYPE
        detection: datatype
```

While it does eventually mention that there's a problem compiling the pattern, it takes a bit to get there. Also, it looks to provide the location hint for the compiled pattern rather than the true location in the source file. 

Ideally, the error would:
- Omit information unrelated to it being a pattern parsing error (or at least prioritize the cause, rather than have it nested within all the problems it caused)
- use the yml file's line and column numbers
- use the yml files erroneous line instead of CurioVar1
- probably shouldn't be JSON output

elsapet:
Closing as partially addressed. I have added some notes to the existing validation issue to include rule pattern validation: https://github.com/Bearer/bearer/issues/429

PR #791, opened 2023-03-14
fix: improve error messaging for custom detector failures

elsapet:
## Description
* Assume a custom detector failure is an external rule and improve error messaging appropriately
* Retain original error message as a debug message

Before
```
{"level":"fatal","time":"2023-03-14T18:17:58+02:00","message":"Failed to start bearer, error with your configuration classifier error: failed to instantiate composition ruby:failed to create detector customDetector:my_rule: error compiling pattern: failed to build: error parsing pattern at 1:5: $4"}
```

After

```
{"level":"fatal","time":"2023-03-14T17:59:12+02:00","message":"could not parse rule my_rule. Is this a custom rule? See documentation on rule patterns and format https://docs.bearer.com/guides/custom-rule/"}
```

Partially addresses #545

The more sophisticated solution to the above, however, would be a YAML validator and a rule pattern validator, so that we can catch any issues when loading the rules, rather than during the scan itself. 

## Checklist

### DIFF
--- a/pkg/commands/process/balancer/process.go
+++ b/pkg/commands/process/balancer/process.go
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"runtime/debug"
 	"strconv"
+	"strings"
 	"time"
 
 	config "github.com/bearer/bearer/pkg/commands/process/settings"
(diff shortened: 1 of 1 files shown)


elsapet:
To move away from JSON output, I tried `zerrors.Errorf()` and a few other things, but these resulted in the error being raised twice. Any suggestions? 

```
Loading rules
Error: filesystem scan error: could not parse rule my_rule. Is this a custom rule? See documentation on rule patterns and format https://docs.bearer.com/guides/custom-rule/
filesystem scan error: could not parse rule my_rule. Is this a custom rule? See documentation on rule patterns and format https://docs.bearer.com/guides/custom-rule/
```

cfabianski (approved):

cfabianski:
What do you think @markmichon ?

markmichon:
@cfabianski messaging lgtm

PR #794, opened 2023-03-15
feat: skip rules with comments in code

### LINKED ISSUE #788: Skip rules with code comments
Add functionality to skip rules based on comments in the code. The `skip-rule` flag allows us to skip a rule entirely. However, we may want something more acute; for example, we may want to skip a rule just for a specific line or section of code. For example: 

```ruby
def my_func
 # bearer:disable ruby_rails_logger
  Rails.logger(current_user.email)
end
```

or 

```ruby
# bearer:disable insecure_net_http
Net::HTTP.start("http://my.api.com/users/search") do
  ...
end
```

elsapet:
## Description

Support comments to disable rules by ID for specific pieces of code. 

When initializing the evaluator, we iterate the tree to read any `bearer:disable` comments and map the given rule IDs with the associated nodes. 

Note for this first iteration, we keep things simple and only support comments above the affected line (or block or method). 

For example: 

```
# bearer:disable ruby_lang_logger, ruby_lang_http_insecure
Net::HTTP.start("http://my.api.com/users/search") do
  logger.warn("Searching for #{current_user.email}")
  ... 
end
```

Closes #788

## Checklist

### DIFF
--- a/new/detector/implementation/custom/custom.go
+++ b/new/detector/implementation/custom/custom.go
@@ -58,10 +58,6 @@ func (detector *customDetector) DetectAt(
 ) ([]interface{}, error) {
 	var detectionsData []interface{}
 
-	if evaluator.RuleDisabledForNode(detector.Name(), node) {
-		return detectionsData, nil
-	}
-
 	for _, pattern := range detector.patterns {
 		results, err := pattern.Query.MatchAt(node)
 		if err != nil {
--- a/new/detector/implementation/custom/custom.go
+++ b/new/detector/implementation/custom/custom.go
@@ -58,6 +58,10 @@ func (detector *customDetector) DetectAt(
 ) ([]interface{}, error) {
 	var detectionsData []interface{}
 
+	if evaluator.RuleDisabledForNode(detector.Name(), node) {
+		return detectionsData, nil
+	}
+
 	for _, pattern := range detector.patterns {
 		results, err := pattern.Query.MatchAt(node)
 		if err != nil {
--- a/new/detector/evaluator/evaluator.go
+++ b/new/detector/evaluator/evaluator.go
@@ -12,11 +12,12 @@ import (
 )
 
 type evaluator struct {
-	lang               languagetypes.Language
-	detectorSet        types.DetectorSet
-	detectionCache     map[langtree.NodeID]map[string][]*types.Detection
-	executingDetectors map[langtree.NodeID][]string
-	fileName           string
+	lang                  languagetypes.Language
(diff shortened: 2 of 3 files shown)


new/language/patternquery/builder/builder.go
```diff
@@ -107,13 +109,53 @@ func (builder *builder) build(rootNode *tree.Node) (*Result, error) {
...
 	}, nil
 }
 
+func (builder *builder) skipRulesForNodes(rootNode *tree.Node) map[string][]*tree.Node {
+	var res map[string][]*tree.Node
+	var rulesSkippedForNode []string
+	// walk tree
+	err := rootNode.Walk(func(node *tree.Node, visitChildren func() error) error {
+		log.Error().Msgf("node type %s", node.Type())
+		if node.Type() == "comment" {
```
elsapet, line 126: Think I'm doing something wrong here. It seems none of the nodes are "comment" types. Maybe `Walk()` isn't what we want here?

new/language/patternquery/builder/builder.go
```diff
@@ -107,13 +109,53 @@ func (builder *builder) build(rootNode *tree.Node) (*Result, error) {
...
-		ParamToContent:  builder.paramToContent,
+		Query:             builder.stringBuilder.String(),
+		ParamToVariable:   paramToVariable,
+		EqualParams:       equalParams,
+		ParamToContent:    builder.paramToContent,
+		SkipRulesForNodes: builder.skipRulesForNodes(rootNode),
 	}, nil
 }
 
+func (builder *builder) skipRulesForNodes(rootNode *tree.Node) map[string][]*tree.Node {
```
didroe: This logic should be applied to the source code and not the patterns. Rather than pattern query builder, you want to do it here:

https://github.com/Bearer/bearer/blob/675ed2b41c97c701d75d689771309370ad8f97e2/new/detector/composition/javascript/javascript.go#L183

(and similar for Ruby)

new/detector/implementation/custom/custom.go
```diff
@@ -58,6 +58,10 @@ func (detector *customDetector) DetectAt(
 ) ([]interface{}, error) {
 	var detectionsData []interface{}
 
+	if evaluator.RuleDisabledForNode(detector.Name(), node) {
```
didroe: WDYT about moving this logic to `Evaluator.detectAtNode`?

didroe (approved):

cfabianski (approved):
Good job @elsapet

PR #795, opened 2023-03-15
feat: add typescript

### LINKED ISSUE #543: Add TS Support
Cross-compile JS rules to TypeScript

vjerci:
## Description
Todo:
- [x] Add typescript as a language
- [x] Add support for method properties
- [x] Remove javascript parser (typescript should cover both js and ts)
- [x] fix class queries
- [x] skip inline type definitions
- [x] basic support for js, ts, jsx and tsx.

fix rules:
- [x] TestJavascriptLangException
- [x] Fix argument detector
    - [x] TestJavascriptReactDangerouslySetInnerHTML
    - [x]  TestJavascriptAwsLambdaSqlInjection
    - [x] TestJavascriptThirdPartySentry
    - [x] TestJavascriptExpressSqlInjection
    - [x] TestJavascriptJWTWeakEncryption

## Related
- close https://github.com/Bearer/bearer/issues/543

## Checklist


new/language/implementation/implementation.go
```diff
@@ -107,6 +107,18 @@ type Implementation interface {
...
+	//
+	// eg. given the following tree sitter
+	// arguments
+	//	formal_parameters
+	//		required_parameter
+	//			identifier
+	//
+	//  if you want to get only identifier instead of required parameter ShouldSkipNode should return true
+	//  for required parameter
+	ShouldSkipNode(node *tree.Node) bool
```
didroe, line 121: I think this would be better inverted. ie. `IsPatternVariableNodeType`.

For JS/TS we should just need `identifier` and for Ruby `identifier` and `constant`.

didroe (approved):

new/language/implementation/implementation.go
vjerci: I've tried that and got multiple ruby and javascript rules failures.
I find it simpler to define a single node type we want to skip than to define 3+ node types we want to parse as variables.

didroe: That's interesting that we got failures.

It really depends on your definition of "simpler". I think it's harder to understand why there's an exception than that variables are parsed as specific node types.

(I'm happy either way really though)

PR #797, opened 2023-03-16
feat(rules): update trigger attributes

### LINKED ISSUE #744: Improve Trigger in rule to better deal with DataType
We currently support 4 different types for `trigger` when writing a rule (`presence`, `absence`, `local` and `global`).

**current**

```yaml
# rule.yml
trigger: local | global | presence | absence
```

## History

We started with only 2 triggers `global` and `local`. So we were firing the rule when we had the patterns matching and we detected some data types either globally or locally.
We then introduced the new trigger `presence` to fire no matter if we had detected some data types or not.

The `presence` is very close to the `global` trigger (especially since we refactored the severity calculation), expect that a `presence` rule can fire even when there are no data types detected globally.

Later on, we introduced the `absence` trigger. This is when you have an auxiliary pattern that's matching but the whole global pattern doesn't. Again, that's not taking into account the data types.

In the end, it feels like we are mixing up 2 things:
- when do we trigger? (`absence` or `presence`)
- should we care about the data types at all? if so, locally or globally?

## Proposal

We could make it a bit easier to read by making the following changes

**proposal**

```yaml
trigger:
  pattern: present | absent # or missing instead of `absent`
  pattern_initializer: <name-of-auxiliary-rule> # would be used in conjunction with `absent` to indicate what is the minimal setup to fire the rule in case the pattern is missing. Replaces `trigger_rule_on_presence_of`
  data_types: global # or not present
```

The `local` use case could be deducted from the rule itself. If we have in the filters `data_types` as a detection we would consider the local scope
A missing `data_types` node, would indicate that we wouldn't take the data types into account to fire the rule. 

## References

- https://github.com/Bearer/bearer/pull/727

elsapet:
## Description

Improve rule trigger. Defaults are as follows

```yaml
trigger:
  match_on: presence
  data_types_required: false
  required_detection: nil 
```

Note, along with these changes, we update the ruby bugsnag snapshot - the update seems like what we should expect to me, but open to input here 🙏 

Closes #744 

## Checklist

### DIFF
--- a/docs/guides/custom-rule.md
+++ b/docs/guides/custom-rule.md
@@ -18,11 +18,14 @@ To better understand the structure of a rule file, let’s look at each key:
 
 - `patterns`: See the section below for the Pattern Syntax.
 - `languages`: An array of the languages the rule applies to. Available values are: `ruby`, `javascript`
-- `trigger`: Defines when the rule should raise a failure. There are four trigger types:
-  - `local`: Use this trigger when your rule directly relies on data type detections in the pattern. Some examples are sending data to a logger, or making an HTTP request that includes sensitive data.
-  - `global`: Some rules don’t match code with a data type directly, but you want them to trigger if Bearer finds any sensitive data types in the project. One example is password strength, where the rule only triggers if sensitive data types are found in the application.
-  - `presence`: Use this trigger when your rule isn’t related to a [data type](/reference/datatypes) detection but on the presence of a pattern. Examples include best practices such as configuration settings like forcing SSL communication.
-  - `absence`: Use this trigger when your rule isn’t related to a [data type](/reference/datatypes) detection but on the absence of a pattern (if we have been able to confirm the presence of an auxiliary pattern). Examples include best practices such as missing configuration like forcing SSL communication.
+- `trigger`: Defines under which conditions the rule should raise a failure. Optional.
+  - `match_on`: Refers to the rule's pattern matches.
+    - `presence`: Default. Examples include best practices such as configuration settings like forcing SSL communication.
+    - `absence`: Rule triggers on the absence of a pattern, but the presence of a `required_detection`. Examples include best practices such as missing configuration like forcing SSL communication. Note: rules that match on `absence` need a `required_detection` to be set.
+  - `required_detection`:  Used with the `match_on: absence` trigger. Indicates which rule is required to activate the failure on the absence of the main rule.
+  - `data_types_required`: Sometimes we may want a rule to trigger only for applications that process sensitive data. One example is password strength, where the rule only triggers if sensitive data types are found in the application.
+    - `false`: Default. Rule triggers whether or not any data types have been detected in the application.
+    - `true`: Rule only triggers if at least one data type is detected in the application.
 - `severity`: This sets the lowest severity level of the rule, by default at `low`. The severity level can automatically increase based on the data type categories (PHI, PD, PDS, PII) detected depending on the rule `trigger` type. A severity level of `warning`, however, will never increase. Bearer groups rule failures by severity, and you can configure the summary report to only fail on specific severity thresholds. Severity is set for each data type group, each of which takes a severity level of `warning`, `low`, `medium`, `high`, or `critical`. A severity level of `warning` won’t cause CI to fail.
 - `metadata`: Rule metadata is used for output to the summary report, and documentation for the internal rules.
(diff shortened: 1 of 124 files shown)


elsapet:
Ping @markmichon for the suggested doc updates on this one 🙏 Issue #744 and comments have further context, if needed

docs/guides/custom-rule.md
```diff
@@ -18,11 +18,14 @@ To better understand the structure of a rule file, let’s look at each key:
...
-  - `global`: Some rules don’t match code with a data type directly, but you want them to trigger if Bearer finds any sensitive data types in the project. One example is password strength, where the rule only triggers if sensitive data types are found in the application.
-  - `presence`: Use this trigger when your rule isn’t related to a [data type](/reference/datatypes) detection but on the presence of a pattern. Examples include best practices such as configuration settings like forcing SSL communication.
-  - `absence`: Use this trigger when your rule isn’t related to a [data type](/reference/datatypes) detection but on the absence of a pattern (if we have been able to confirm the presence of an auxiliary pattern). Examples include best practices such as missing configuration like forcing SSL communication.
+- `trigger`: Defines under which conditions the rule should raise a failure. Optional.
+  - `match_on`: Refers to the rule's pattern matches.
+    - `presence`: Default. Examples include best practices such as configuration settings like forcing SSL communication.
```
markmichon:
```suggestion
    - `presence`: Triggers if the rule's pattern is detected. (Default)
```
I'm not sure the best practice example makes sense for presence, as we don't tend to notify when 'good' things happen.

docs/guides/custom-rule.md
```diff
@@ -18,11 +18,14 @@ To better understand the structure of a rule file, let’s look at each key:
...
-  - `local`: Use this trigger when your rule directly relies on data type detections in the pattern. Some examples are sending data to a logger, or making an HTTP request that includes sensitive data.
-  - `global`: Some rules don’t match code with a data type directly, but you want them to trigger if Bearer finds any sensitive data types in the project. One example is password strength, where the rule only triggers if sensitive data types are found in the application.
-  - `presence`: Use this trigger when your rule isn’t related to a [data type](/reference/datatypes) detection but on the presence of a pattern. Examples include best practices such as configuration settings like forcing SSL communication.
-  - `absence`: Use this trigger when your rule isn’t related to a [data type](/reference/datatypes) detection but on the absence of a pattern (if we have been able to confirm the presence of an auxiliary pattern). Examples include best practices such as missing configuration like forcing SSL communication.
+- `trigger`: Defines under which conditions the rule should raise a failure. Optional.
```
markmichon, line 21: This file also needs the trigger key removed from the 'starter file' at the end of the document.

markmichon (changes requested):

pkg/commands/process/settings/rules.go
```diff
@@ -176,27 +176,52 @@ func buildRules(definitions map[string]RuleDefinition, enabledRules map[string]s
...
+			}
+		}
+
+		isLocal := false
+		for _, rulePattern := range definition.Patterns {
+			if strings.Contains(rulePattern.Pattern, "$<DATA_TYPE>") {
+				isLocal = true
+				break
+			}
+		}
```
elsapet, line 202: Not sure about this. I need to know, before the rule is run, whether or not it counts as a "local" rule. What I have here feels a bit hack-ish and maybe points to the wider issue that we no longer support rules who have both patterns with `$<DATA_TYPE>` and patterns without `$<DATA_TYPE>` as well as before 🤔

docs/guides/custom-rule.md
elsapet: Good catch - thanks 🙏

markmichon (approved):

pkg/commands/process/settings/rules.go
cfabianski: definitely feels like the rules should be separated but that prevents us from having a level set to warning (when no data types are detected) and that would increase to critical if we detect something (e.g. gender).
WDYT? @gmontard

elsapet: To represent the `trigger: local` situation, we could use the following:

```
- trigger:
      datatypes_required: true
```

I chose against it as it felt a bit superfluous since the patterns for local rules all have `$<DATA_TYPE>` in them anyway, but it would allow us to differentiate these cases (local rules, and presence rules that may or may not have `$<DATA_TYPE>` in their patterns)

elsapet: @cfabianski I've removed the rule split and updated the Rego. Looking forward to your thoughts 🙏

cfabianski (approved):
Great job!

PR #827, opened 2023-03-27
feat: load rules externally

### LINKED ISSUE #799: Load rules from external repository
We currently need to release a new version every time we write a new rule. This is tiresome and prevents quick iteration. Instead, we could move rules (the rule YAML files) to their own repository: https://github.com/Bearer/bearer-rules

This would simplify the way to customize rules as you would need to fork only the rules folder rather than the whole bearer repo. 

Some things for consideration: 
- How would the rule integration tests work? 
- What about the documentation? Right now the documentation website includes the documentation of the rules which uses the metadata from the rule files themselves
- How would this relate (if at all) to the current external rule functionality?
  - We would also no longer need to have an external rule folder defined in the config. Instead you just specify which DB you want to use (by default, [github.com/bearer/bearer-rules](http://github.com/bearer/bearer-rules) would be picked up)
  - However, external rule folder may be solving a different problem - for example, it provides a quick way to add a new rule; whereas, forking the rules repo lets us edit existing rules.
- Should we download all the rules each time or can we do with the subset of languages identified at first.
  -  Ideally, we should list all the languages detected at the beginning and download only the rules for these languages (@cfabianski)

Pending tasks

- [ ] Move documentation URLs to bearer-action rule YAML files https://github.com/Bearer/bearer-rules/issues/6
- [ ] Update README.md for bearer-rules repo
- [ ] Documentation around bearer-rules repo and how it is used

elsapet:
## Description

* Load rules from external repository https://github.com/Bearer/bearer-rules (caching where possible)
* Remove rules directory
* Remove rules/integration tests
* Update e2e tests to use external rules only

Closes #799 

## Checklist


vjerci (approved):

cfabianski:
I couldn't see https://github.com/Bearer/bearer/pull/835 🙈

e2e/rules/.snapshots/TestRubyRailsDefaultEncryptionSchema-testdata-data-ruby_rails_default_encryption_schema_rb
```diff
@@ -13,7 +13,7 @@ warning:
 
             ## Resources
             - [Ruby on Rails Active Record encryption](https://guides.rubyonrails.org/active_record_encryption.html)
-        documentation_url: https://docs.bearer.com/reference/rules/ruby_rails_default_encryption
+        documentation_url: ""
```
didroe, line 16: Is this expected?

pkg/commands/process/settings/rules.go
```diff
@@ -240,3 +275,21 @@ func buildRules(definitions map[string]RuleDefinition, enabledRules map[string]s
...
+}
+
+func cleanupRuleDirFiles(bearerRulesDir string) error {
+	os.RemoveAll(bearerRulesDir)
+	return nil
+}
+
+func bearerRulesDir() string {
+	tmpDir := os.TempDir()
+	if tmpDir == "/tmp" {
```
didroe: Why is the special case needed?

pkg/commands/process/settings/rules.go
```diff
@@ -240,3 +275,21 @@ func buildRules(definitions map[string]RuleDefinition, enabledRules map[string]s
...
+		return "/tmp/bearer-rules/"
+	}
+	return tmpDir + "bearer-rules"
```
didroe: Do we need the path separator here?
```suggestion
	return tmpDir + "/bearer-rules"
```
(or filepath.Join)

didroe (approved):

elsapet:
Thanks @cfabianski - will add it to bearer-rules 🙏  
Also planning to do a full copy / paste of current `main` prior to release to make extra sure all of the rules are up-to-date 👍

e2e/rules/.snapshots/TestRubyRailsDefaultEncryptionSchema-testdata-data-ruby_rails_default_encryption_schema_rb
elsapet: Yes, it's now considered an external rule like the other E2E rules, so its doc URL is not auto-generated any more. Before we were relying on the actual rule in the rules repo.

pkg/commands/process/settings/rules.go
elsapet: Ah, thanks - it's not 🙈 It was from me trying to get the CI to be happy, but we can remove it 👍

pkg/commands/process/settings/rules.go
elsapet: Yeah it's a bit weird. From what I can tell with `os.TempDir()` is that if a temp directory exists, the trailing slash is included (at least for my OS), but if there's no temp directory, it returns `/tmp` without a trailing slash.. 

Maybe best to do a check and add it if it's missing? What do you think?

didroe: It looks like `filepath.Join` will handle the differences:
```go
	fmt.Println(filepath.Join("/tmp", "rules")) //      /tmp/rules
	fmt.Println(filepath.Join("/tmp/", "rules")) //     /tmp/rules
```

elsapet: Ah, brilliant, thanks! 🤓

PR #828, opened 2023-03-28
feat(report): extend saas report

cfabianski:
## Description

Extends SaaS Report and make it available as a report option

## Checklist


gotbadger (approved):
Maybe report cloud would make more sense but not that big of a deal

PR #834, opened 2023-03-29
feat: dataflow deduplication

vjerci:
## Description
This pr sorts out various detectors duplication bugs we had.
Previously in some cases we were reporting duplicatate risk detectors in our dataflow, causing some detections to be ignored.
This pr solves it by reworking risk detectors, meaning schema and internals are changed.
Due to those changes we had to change our policy_risks.rego, which was also simplified.

### Related
https://github.com/Bearer/bearer/issues/819 

## Checklist


pkg/commands/process/settings/rules/javascript/lang/logger/.snapshots/TestJavascriptLangLogger--console.yml
```diff
@@ -28,5 +28,5 @@ high:
         - Personal Data
       parent_line_number: 2
       snippet: console.log("user email is", user.email)
-      fingerprint: 907f568d6d401d24176f4ad514eba094_1
+      fingerprint: 907f568d6d401d24176f4ad514eba094_0
```
cfabianski: Should not really be this order 😕

pkg/commands/process/settings/rules/javascript/lang/weak_encryption.yml
```diff
@@ -42,9 +47,7 @@ auxiliary:
               - createHmac
               - createHash
           - variable: ALGORITHM
-            values:
-              - '"md5"'
-              - '"sha1"'
+            regex: (?i)['"]sha1|md5["']
```
cfabianski, line 50: 👍

cfabianski (commented):
I'm wondering about the ordering of the failures in the snapshots (`<fingerprint_1>_5` shows up before `<fingerprint_1>_0` for instance) but that looks very close to what we want!

elsapet (approved):
Code looks good!

One question though - looking at the snapshot changes, it feels a bit weird to me to have two errors when the line number, parent line number, and snippet are the same. Code-wise, I know it is expected because there are different data types involved, but have these cases been confirmed with product? If not, maybe worth checking 🤔

PR #841, opened 2023-03-30
fix: fix category grouping issue

elsapet:
### The issue
It seems we've had a long-standing issue with data category groups. We were not including parent category groups correctly, when compiling data category groups; namely, data categories such as Communication in the PHI and PII category groups did not include the parent category group (PD) in their overall category groups. This meant that Data types in PHI and PII data category groups did not include PD in their data category groups. 

For example, email data type was incorrectly considered to be in the PII group, but not in the PD group. 

This had an effect on the severity calculation. 

### The fix

In this PR we fix grouping issue, so that now email data type is in both the PII and PD groups.

### The consequences

The main consequence is in severity calculation, which used data type groups (PII, PD). As a result, the main change of this PR is snapshots being updated. 

## Checklist


cfabianski (approved):
Nice!

PR #842, opened 2023-03-30
feat(load rules externally): add disable default rules flag

elsapet:
## Description

Adds a new rules flag to disable default rules

This will be used by the CI on the upcoming bearer-rules repository. This change does two things:
* improves CI performance (because we can avoid downloading rules from GH from the binary)
* makes it more straight-forward to test unreleased rules (because we can load them as external rules without worrying about rule ID conflicts)

Relates to #799

## Checklist

### DIFF
--- a/docs/_data/bearer_scan.yaml
+++ b/docs/_data/bearer_scan.yaml
@@ -13,6 +13,9 @@ options:
     - name: debug
       default_value: "false"
       usage: Enable debug logs
+    - name: disable-default-rules
+      default_value: "false"
+      usage: Disables all default and built-in rules.
     - name: disable-domain-resolution
       default_value: "true"
       usage: |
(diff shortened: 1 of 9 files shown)


docs/_data/bearer_scan.yaml
```diff
@@ -13,6 +13,9 @@ options:
     - name: debug
       default_value: "false"
       usage: Enable debug logs
+    - name: disable-default-rules
+      default_value: "false"
+      usage: Disables all default and built-in rules.
```
elsapet, line 18: Ping @markmichon for the new flag and flag description 🙏 
I'm not sure how much detail we want in the description, but for context when enabled, this flag means a Bearer scan would only run rules in the external rules folder.

markmichon: Let's keep it minimal like you have it now. I can enhance the custom rule guide to better explain it.

vjerci (approved):

PR #860, opened 2023-04-04
docs: external rules logic

markmichon:
## Description
Updates rules documentation to handle the move to an external rules repo. Specific caveats:
- Relies on hard-coded language support to confirm contents are rules. Docs will require updates when new languages are added, so shouldn't be a problem in the near term.
- Expects `bearer/bearer-rules` as the external rules repo for both processing data and linking to rule source.
- Always pulls the latest release. This means a doc rebuild will always pull the latest rules release.
- Somewhat messy caching in docs dev environment. Watch for temp dirs when changing branches until the included gitignore is everywhere. Won't affect production.

Not yet included:
- Messaging related to external rules repo.
- UI for rules version in the `reference/rules` page. Decisions on how to handle this TBD.

## Checklist

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,3 +1,4 @@
 node_modules/
 _site/
+_tmp/
 .cache/
--- a/docs/_data/rules.js
+++ b/docs/_data/rules.js
@@ -1,11 +1,13 @@
 const { readFile, readdir } = require("node:fs/promises");
 const { statSync } = require("fs");
+const fetch = require("node-fetch");
 const path = require("path");
 const yaml = require("js-yaml");
 const cweList = require("./cweList.json");
 const gitly = require("gitly");
 const source = "bearer/bearer-rules";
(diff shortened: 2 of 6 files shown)


elsapet (approved):
Looks good! 

A thought around the hard-coding, we could flip this and hardcode the non-language folders (`scripts`, `.github`) instead. Unlikely we'll need more than these two 🤔  

```
if (isDirectory(dirPath) && !excludedDirectories.includes(dir)) {
 ...
}
```

markmichon:
> Looks good!
> 
> A thought around the hard-coding, we could flip this and hardcode the non-language folders (`scripts`, `.github`) instead. Unlikely we'll need more than these two 🤔
> 
> ```
> if (isDirectory(dirPath) && !excludedDirectories.includes(dir)) {
>  ...
> }
> ```

Makes sense. The latest commit updates logic.

PR #861, opened 2023-04-04
docs: add update instructions to installation page

### LINKED ISSUE #813: Docs: Adding "Upgrade" instruction
It would be helpful to add an "upgrade/update" instruction page in the doc below the "installation" one:
https://docs.bearer.com/reference/installation/

markmichon:
## Description
Updates installation instructions to include update instructions for each installation method.

## Related
- Close #813 

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,12 @@ curl -sfL https://raw.githubusercontent.com/Bearer/bearer/main/contrib/install.s
   ```bash
   brew install bearer/tap/bearer
   ```
+
+  Update an existing installation with the following:
+  
+  ```bash
+  brew update && brew update bearer/tap/bearer
+  ```
 </details>
 
 <details>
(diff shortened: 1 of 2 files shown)


gotbadger:
@markmichon I think the docker one does actually need action in some cases if you are using latest to get the new version you need to do something like

```
docker image pull bearer/bearer:latest-amd64
```

cfabianski:
How about we create a totally new page/section dedicated to update/upgrade?
That way we could link it up with the message when we detect it's not the latest?

markmichon:
@cfabianski I'm more inclined to have it be a dedicated section at the bottom of `installation` and we could link to `installation#update`. Worried about page fatigue. Thoughts?

cfabianski:
> @cfabianski I'm more inclined to have it be a dedicated section at the bottom of `installation` and we could link to `installation#update`. Worried about page fatigue. Thoughts?

new section is good enough yes.
Wanna update this message too with the link? https://github.com/Bearer/bearer/blob/dc35d997c67c1719fae83b4adc77e4f4b7476b6f/pkg/commands/artifact/run.go#L229

markmichon:
> > @cfabianski I'm more inclined to have it be a dedicated section at the bottom of `installation` and we could link to `installation#update`. Worried about page fatigue. Thoughts?
> 
> new section is good enough yes. Wanna update this message too with the link?
> 
> https://github.com/Bearer/bearer/blob/dc35d997c67c1719fae83b4adc77e4f4b7476b6f/pkg/commands/artifact/run.go#L229

@cfabianski we can, but that's out of scope of this PR. Rather not tie upgrade instructions to a future version release. I'll create a new issue for that.

cfabianski (approved):

PR #862, opened 2023-04-05
chore(docker): update goreleaser for docker image

cfabianski:
## Description

It feels a bit weird to add the suffix `-amd64` while there is only 1 architecture there.
That would simplify the documentation as well. 

## Checklist


gotbadger (changes requested):
PR needs to contain the documentation changes before merging.

gotbadger (approved):
wont update doc but just offer this as simpler approach

PR #865, opened 2023-04-05
feat: handle same line number deduplication

### LINKED ISSUE #847: Deduplicate same detector vulnerabilities that have the same line number
For a simple rule like this:
```
console.log(...$<DATA_TYPE>...)
```
and for code example
```
console.log(user.name, user.gender)
```

We have 2 vulnerability detections that from the user's perspective look like a bug.
The one is for `user.name` and the second one is for `user.gender`

In such cases, the expected behavior is to report only the one with the highest security vulnerability rating and remove others from the security report.

vjerci:
## Description
We want to deduplicate same line number detections by same detector.
This pr handles it in a way that it takes a single highest severity detection and ignores the rest detections of same detector that have the same line number.

## Related
- closes https://github.com/Bearer/bearer/issues/847

## Checklist


pkg/report/output/security/security.go
```diff
@@ -534,7 +542,92 @@ func iterativeDigitsCount(number int) int {
 	return count
 }
 
-func SortResult(data []Result) {
+// removeDuplicates removes detections for same detector with same line number by keeping only a single highest severity detection
+func removeDuplicates(data map[string][]Result) map[string][]Result {
```
didroe, line 547: I wonder if this could be simplified a lot by just using a map that stores whether a given rule id and line number have been seen before. Then if you process things in descending severity then "seen before" must be at least the same severity. WDYT?

didroe (approved):
I realise this isn't a decision you've taken but I wonder if this will be annoying for users. As you could "fix" an issue but it doesn't actually fix it as you only dealt with 1 out of many occurrences. Let's see

pkg/report/output/security/security.go
vjerci: I don't think it would simplify it, I've though about it (going descending way) but having some sort of a nested  map `map[string][string][string]bool`  - with keys being rule_id, filename, line_number imo doesn't scale quite well code wise. Especially if further down the line you want to embed more logic into it (eg. if parents are the same content and line number) ending up with more map levels.

didroe: You can use a struct as the key: https://go.dev/play/p/LudMcVBoQ3R

vjerci: nice, will make the change

cfabianski:
> I realise this isn't a decision you've taken but I wonder if this will be annoying for users. As you could "fix" an issue but it doesn't actually fix it as you only dealt with 1 out of many occurrences. Let's see

I can see pros and cons to both. 

- What we’re checking is that there is a leakage of sensitive data in a specific logger call, not really how many. The remediation is the same, it’s the same line, same CWE, same logger instance.
- If we put out two, when you look at the output, you won’t understand why you have 100% the same and you’d directly think it’s a duplicate/bug.

Let's see yeah

PR #888, opened 2023-04-11
chore: better handle typescript

cfabianski:
## Description

Typescript is a bit special so we should handle it as such.

## Checklist

### DIFF
--- a/pkg/commands/artifact/run.go
+++ b/pkg/commands/artifact/run.go
@@ -17,6 +17,7 @@ import (
 	"github.com/hhatto/gocloc"
 	"github.com/rs/zerolog/log"
 
+	"golang.org/x/exp/maps"
 	"golang.org/x/xerrors"
 
 	"github.com/bearer/bearer/cmd/bearer/build"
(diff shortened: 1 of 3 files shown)


vjerci (approved):

pkg/commands/artifact/run.go
```diff
@@ -215,7 +216,7 @@ func (r *runner) scanArtifact(ctx context.Context, opts flag.Options) (types.Rep
 // Run performs artifact scanning
 func Run(ctx context.Context, opts flag.Options, targetKind TargetKind) (err error) {
 	if !opts.Quiet {
-		outputhandler.StdErrLogger().Msg("Loading rules")
+		output.StdErrLogger().Msg("Loading rules")
```
elsapet, line 219: Curious why we can't keep `outputhandler` here? Doesn't it mean we are importing `pkg/util/output` twice?

pkg/commands/artifact/run.go
```diff
@@ -252,16 +248,6 @@ func Run(ctx context.Context, opts flag.Options, targetKind TargetKind) (err err
...
 
-	if !r.CacheUsed() && scanSettings.CacheUsed {
-		// re-cache rules
-		if opts.ScanOptions.Force && !opts.ScanOptions.Quiet {
-			outputhandler.StdOutLogger().Msgf("Caching rules")
-		}
-		if err = settings.RefreshRules(scanSettings, opts.ExternalRuleDir, opts.RuleOptions, formatFoundLanguages(inputgocloc.Languages)); err != nil {
-			return err
-		}
-	}
```
elsapet: Not sure I understand the changes in this file. This is removing the work I did to cache the bearer-rules. Is it a git / merge issue?

elsapet (changes requested):
Not sure I'm understanding the pkg/commands/artifact/run.go changes - maybe a manual rebase before merging?

elsapet (approved):

PR #891, opened 2023-04-12
feat: support spread element in JS/TS

cfabianski:
## Description

Add support for `spread_element` so that data type nested within spread_element could be found

e.g. 

```javascript
const sensitiveInfo = { firstName: "John", lastName: "Doe" }
const user = { ...sensitiveInfo, foo: "Bar" }
logger.info(user)
```

## Checklist


didroe (approved):
Nice!

PR #905, opened 2023-04-14
docs: replacing bear-publishing examples by OWASP Juice Shop

gmontard:
## Description
replacing bear-publishing example in the doc by OWASP Juice Shop

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,16 +157,16 @@ curl -sfL https://raw.githubusercontent.com/Bearer/bearer/main/contrib/install.s
 
 ### Scan your project
 
-The easiest way to try out Bearer is with our example project, [Bear Publishing](https://github.com/Bearer/bear-publishing). It simulates a realistic Ruby application with common security flaws. Clone or download it to a convenient location to get started.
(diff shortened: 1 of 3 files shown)


markmichon (commented):
@gmontard Report page still has mentions of bear-publishing and ruby output for dataflow report. Otherwise lgtm.

gmontard:
> @gmontard Report page still has mentions of bear-publishing and ruby output for dataflow report. Otherwise lgtm.

@markmichon Yes I was thinking about changing it but I won't be able to mention the part of where we found the data in the database because there is no schema in this App.

So I'm wondering, is it better to change it and remove this part, or leave it with Bear Publishing as an exemple?

markmichon:
> > @gmontard Report page still has mentions of bear-publishing and ruby output for dataflow report. Otherwise lgtm.
> 
> @markmichon Yes I was thinking about changing it but I won't be able to mention the part of where we found the data in the database because there is no schema in this App.
> 
> So I'm wondering, is it better to change it and remove this part, or leave it with Bear Publishing as an exemple?

@gmontard ah, that makes total sense. Yeah let's leave it then. Approving.

markmichon (approved):

PR #906, opened 2023-04-15
docs: add variable joining to rule guide

### LINKED ISSUE #858: Document new pattern variable behaviour
Document the new functionality with pattern variables introduced in https://github.com/Bearer/bearer/pull/856.

The new behaviour is that, when a parent rule uses another by referring to it using a filter, variables with the same names in patterns of both rules will now always refer to the same code location (AST node really). 

TODO: example of how that can be useful.

markmichon:
## Description
Explains the "variable joining" that can result from specific uses of auxiliary rules.

## Related
- close #858 

## Checklist

### DIFF
--- a/docs/guides/custom-rule.md
+++ b/docs/guides/custom-rule.md
@@ -34,7 +34,7 @@ To better understand the structure of a rule file, let’s look at each key:
   - `associated_recipe`: Links the rule to a [recipe]({{meta.sourcePath}}/tree/main/pkg/classification/db/recipes). Useful for associating a rule with a third party. Example: “Sentry” (Optional)
   - `remediation_message`: Used for internal rules, this builds the documentation page for a rule. (Optional)
   - `documentation_url`: Used to pass custom documentation URL for the summary report. This can be useful for linking to your own internal documentation or policies. By default, all rules in the main repo will automatically generate a link to the rule on [docs.bearer.com](/). (Optional)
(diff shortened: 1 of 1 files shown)


markmichon:
@didroe Please take a look when you have some time. I think it stays true to your initial explanation, but if not don't hesitate to revise or leave a note and I'll revisit it when I return. 

I think there's an opportunity to improve this in the future, but it may benefit from more rules that bump into this case in order to learn more about potential frictions.

docs/guides/custom-rule.md
```diff
@@ -236,6 +236,81 @@ metadata:
...
+severity: medium
+metadata:
+  description: "Variable joining example"
+  remediation_message: ""
+  cwe_id:
+    - 601
+  id: "my_rule"
+```
+
+The `patterns` portion at the beginning should look familiar. The only difference compared to most rules is that it references `detection` types that are auxiliary rules. The main pattern looks for `sql_query`, and then uses filters to tell Bearer CLI to apply each detection to any code it finds in `$<QUERY>`. In this case, it wants to trigger the rule using `my_rule_user_input` but NOT `my_rule_sanitized`. I've labeled the core rule/pattern combo as rule 0, then each Aux rule as rules 1, 2, and 3 to make them easier to follow.
```
didroe, line 297: > **I've** labeled the [...]

Should we be using first person here?

didroe (approved):

Issue #895, opened 2023-04-12
Improved Battle Tests


gotbadger:
Our original battle tests focused on testing a wide array of projects to try and find breaking bases in unusual codebases. Now we want to focus on some specific projects that will demonstrate how bearer works and performs and more importantly give us a baseline for how rules and detection improvements lead to overall better outcomes.

## Projects
We will want to add more to this list as new languages are added or we found new interesting projects. Initial list looks like this:

- https://github.com/Bearer/juice-shop (TS)
- https://github.com/Bearer/railsgoat (Ruby)
- https://github.com/Bearer/WebGoat (Java)
- https://github.com/Bearer/NodeGoat (JS)
- https://github.com/forem/forem (JS)
- https://github.com/freeCodeCamp/chapter (TS)
- https://github.com/mastodon/mastodon (Ruby)
- https://github.com/frab/frab (Ruby)
- https://github.com/discourse/discourse (Ruby)
- https://github.com/diaspora/diaspora (Ruby)
- https://github.com/TryGhost/Ghost (JS)
- https://github.com/wekan/wekan (JS)
- https://gitlab.com/gitlab-org/gitlab (Ruby)
- https://github.com/grafana/grafana (TS)

## Tasks

- [x] Investigate how to keep forked repos up-to-date
  - Not straight-forward. Can we just use non-forked original repos?
- [x] Set up Bearer Cloud accounts @elsapet
- [x] Add rule version to metadata that is sent during bearer scan @elsapet https://github.com/Bearer/bearer/pull/913
- [x] Create job to kick off scan process @didroe https://github.com/Bearer/bearer/pull/897
    - Q: Do we need forked repos? 
- [x] Get scan data into Bearer Cloud account
- [x] Add new repositories to workflow @elsapet https://github.com/Bearer/bearer/pull/923
- [x] Set up non-testing Bearer Cloud account on Prod (Battletest) @elsapet 
- [ ] Use canary builds?

elsapet:
@gotbadger would we want an option to run the battle tests on canary builds?

gotbadger:
@elsapet no I dont think it would fit in with our review processes anyway.

PR #913, opened 2023-04-18
feat(kpi): add bearer and bearer rules version to client meta

elsapet:
## Description
Add Bearer version and Bearer Rules version to meta we send to Client
This will be used to allow us to select specific Bearer and Bearer Rules versions for the KPI reports

Relates to #895

## Checklist

### DIFF
--- a/pkg/commands/process/settings/rules.go
+++ b/pkg/commands/process/settings/rules.go
@@ -18,18 +18,30 @@ var (
 )
 
 func RefreshRules(config Config, externalRuleDirs []string, options flag.RuleOptions, foundLanguages []string) (err error) {
-	builtInRules, rules, _, err := loadRules(externalRuleDirs, options, foundLanguages, true)
+	builtInRules, rules, _, bearerRulesVersion, err := loadRules(externalRuleDirs, options, foundLanguages, true)
 	config.BuiltInRules = builtInRules
 	config.Rules = rules
+	config.BearerRulesVersion = bearerRulesVersion
 
 	return
 }
 
-func loadRules(externalRuleDirs []string, options flag.RuleOptions, foundLanguages []string, force bool) (map[string]*Rule, map[string]*Rule, bool, error) {
+func loadRules(
+	externalRuleDirs []string,
+	options flag.RuleOptions,
+	foundLanguages []string,
+	force bool) (
+	map[string]*Rule,
+	map[string]*Rule,
+	bool,
+	string,
+	error,
+) {
 	definitions := make(map[string]RuleDefinition)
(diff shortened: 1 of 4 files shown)


pkg/commands/process/settings/rules.go
```diff
@@ -18,18 +18,30 @@ var (
 )
 
 func RefreshRules(config Config, externalRuleDirs []string, options flag.RuleOptions, foundLanguages []string) (err error) {
-	builtInRules, rules, _, err := loadRules(externalRuleDirs, options, foundLanguages, true)
+	builtInRules, rules, _, bearerRulesVersion, err := loadRules(externalRuleDirs, options, foundLanguages, true)
```
elsapet: `loadRules` is now returning a lot of things; what's the best practice here?

pkg/commands/process/settings/settings.go
```diff
@@ -40,15 +40,16 @@ type WorkerOptions struct {
...
+	Client             *api.API
+	Worker             WorkerOptions      `mapstructure:"worker" json:"worker" yaml:"worker"`
+	Scan               flag.ScanOptions   `mapstructure:"scan" json:"scan" yaml:"scan"`
+	Report             flag.ReportOptions `mapstructure:"report" json:"report" yaml:"report"`
+	Policies           map[string]*Policy `mapstructure:"policies" json:"policies" yaml:"policies"`
+	Target             string             `mapstructure:"target" json:"target" yaml:"target"`
+	Rules              map[string]*Rule   `mapstructure:"rules" json:"rules" yaml:"rules"`
+	BuiltInRules       map[string]*Rule   `mapstructure:"built_in_rules" json:"built_in_rules" yaml:"built_in_rules"`
+	CacheUsed          bool               `mapstructure:"cache_used" json:"cache_used" yaml:"cache_used"`
+	BearerRulesVersion string             `mapstructure:"bearer_rules_version" json:"bearer_rules_version" yaml:"bearer_rules_version"`
```
elsapet, line 52: Nervous of making `Config` a hold-all... but I couldn't think of a better place for this to go. I considered adding it to `flag.ScanOptions` or `flag.ReportOptions` but it's not strictly an option so this didn't seem right. Suggestions welcome 🙏

pkg/commands/process/settings/rules.go
didroe: It might be better to return a struct for the non-error values.

pkg/commands/process/settings/settings.go
didroe: I think this is ok for now

didroe (approved):

PR #916, opened 2023-04-18
fix: rule caching and language support

### LINKED ISSUE #914: Rules are always downloaded if repo contains non-rule-supported file types (HTML, CSS, txt)
## Description & Reproduction

We are always refreshing the rules and never relying on the cached version. This is a result of the fix here: https://github.com/Bearer/bearer/pull/904

Found languages can include a variety e.g. CSS, HTML, plain text and as such the logic for the above fix is too naïve. We need to ask if we have rules for the found language. 

## Repro steps

1. Scan a repository that has a variety of file types e.g. `.rb`, `.txt`, `.html`
2. Re-scan the same repository

## Expected Behavior

On the re-scan we should be using cached rules

## Actual Behavior

We download the rules every time

## Possible Fix

Could possibly update the [`anySupportedLanguagesPresent` method](https://github.com/Bearer/bearer/blob/main/pkg/commands/artifact/run.go#L392) to return the found and supported languages, and use this in rules cache process

## Your Environment

- Operating System and version: running locally 
- Output of 'bearer version':
```
bearer version: dev / 
sha: cad5fcdecf81921f2776641c19de558a844a0d7f
```

elsapet:
## Description

* Extract support languages to rules file
* Fix logic in rules caching - check found language has rule support before we re-download the rules 🙈 

Closes #914

## Checklist

### DIFF
--- a/pkg/commands/process/settings/rules.go
+++ b/pkg/commands/process/settings/rules.go
@@ -17,6 +17,14 @@ var (
 	defaultAuxiliaryRuleType = "verifier"
 )
 
+func GetSupportedRuleLanguages() map[string]bool {
+	return map[string]bool{
+		"ruby":       true,
+		"javascript": true,
+		"typescript": true,
+	}
+}
+
 func RefreshRules(config Config, externalRuleDirs []string, options flag.RuleOptions, foundLanguages []string) (err error) {
 	result, err := loadRules(externalRuleDirs, options, foundLanguages, true)
 	config.BuiltInRules = result.BuiltInRules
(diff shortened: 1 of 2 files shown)


pkg/commands/process/settings/rules.go
```diff
@@ -17,6 +17,14 @@ var (
 	defaultAuxiliaryRuleType = "verifier"
 )
 
+func GetSupportedRuleLanguages() map[string]bool {
+	return map[string]bool{
+		"ruby":       true,
+		"javascript": true,
+		"typescript": true,
+	}
+}
```
elsapet, line 26: Hard coding this for speed: I'm wanting to know which languages have rule support without doing any API calls e.g. to get the folder names from bearer-rules. 

Not sure this solution is ideal - am open to suggestions 🙏

didroe (approved):

PR #919, opened 2023-04-19
feat: java language support

vjerci:
## Description
This pr adds basic java support.
It adds string detector, object detector, as well as flow analysis support for java.

This pr as well adds support for integration testing rules with certain compositions.

## Related
https://github.com/Bearer/bearer/issues/894

## Checklist


new/detector/implementation/java/object/object.go
```diff
@@ -0,0 +1,194 @@
...
+	results, err := detector.classNameQuery.MatchAt(node)
+	if len(results) == 0 || err != nil {
+		return nil, err
+	}
+
+	result := results[0]
+
+	className := result["class_name"].Content()
+
+	properties, err := generic.GetNonVirtualObjects(evaluator, result["right"])
```
didroe: Does `result["right"]` exist?

new/detector/implementation/java/object/object.go
```diff
@@ -0,0 +1,194 @@
...
+	fieldAcessQuery, err := lang.CompileQuery(`(field_access object: (_) @object field: (identifier) @field) @root`)
+	if err != nil {
+		return nil, fmt.Errorf("error compiling call query: %s", err)
+	}
+
+	return &objectDetector{
+		assignmentQuery:    assignmentQuery,
+		classNameQuery:     classNameQuery,
+		classPropertyQuery: classPropertyQuery,
+		fieldAcessQuery:    fieldAcessQuery,
```
didroe:
```suggestion
		fieldAccessQuery:    fieldAccessQuery,
```

new/detector/implementation/java/ruby_test.go
```diff
@@ -0,0 +1,23 @@
+package ruby_test
```
didroe:
```suggestion
package java_test
```
(also filename and test names)

new/detector/implementation/java/string/string.go
```diff
@@ -0,0 +1,36 @@
...
+
+func (detector *stringDetector) Name() string {
+	return "string"
+}
+
+func (detector *stringDetector) DetectAt(
+	node *tree.Node,
+	evaluator types.Evaluator,
+) ([]interface{}, error) {
+	if node.Type() == "string_literal" {
```
didroe, line 27: We could also support `+`

new/language/implementation/java/java.go
```diff
@@ -0,0 +1,308 @@
...
+		// public class Main {
+		//
+		// }
+		case "class_body":
+			previousScope := scope
+			scope = implementation.NewScope(previousScope)
+			err := visitChildren()
+			scope = previousScope
+			return err
+			// public class Main {
```
didroe:
```suggestion
		// public class Main {
```

new/language/implementation/java/java.go
```diff
@@ -0,0 +1,308 @@
...
+}
+
+func (javaImplementation *javaImplementation) ShouldSkipNode(node *tree.Node) bool {
+	return false
+}
+
+func (*javaImplementation) PatternLeafContentTypes() []string {
+	return []string{
+		// todo: see if type identifier should be removed from here (User user) `User` is type
+		// todo see if `modifier` should be here it's a single token containing `string public final static`
```
didroe: - Identifier is good I think
- Yes I'd say add `modifier`, though we probably won't use it in practice

didroe (approved):

new/detector/implementation/java/object/object.go
vjerci: yep

new/detector/implementation/java/string/string.go
vjerci: lets see if we need it.

new/detector/implementation/java/object/object.go
```diff
@@ -0,0 +1,194 @@
...
+	return objects, nil
+}
+
+func (detector *objectDetector) getClassName(node *tree.Node, evaluator types.Evaluator) ([]interface{}, error) {
+	results, err := detector.classNameQuery.MatchAt(node)
+	if len(results) == 0 || err != nil {
+		return nil, err
+	}
+
+	result := results[0]
```
didroe: If you're only expecting a single result you can use `MatchOnceAt` instead.

new/detector/implementation/java/object/object.go
didroe: I don't see it here: 
https://github.com/Bearer/bearer/pull/919/files#diff-f2858e7bb9c4cf328b23d5a8c3543c6d51b3819021dddda306e52944026c870bR36-R39

vjerci: my bad thought this one is about assigmentQuery, will fix it

new/detector/implementation/java/object/object.go
vjerci: shouldn't that method be called `getSingleResult`? and the other one probably `getAllNestedResults`? `matchOnceAt` doesn't seems explicit but implicit.

vjerci: Fixed, I'm resolving this, and merging, However, I still think that the method naming there is a bit confusing.

PR #932, opened 2023-04-24
build: use ubuntu as base of docker image

didroe:
## Description

Use Ubuntu instead of Alpine for the base of the docker image. 

Now we're dynamically linking against libc, the current Alpine image no longer works. I tried all three compatibility libraries mentioned [here](https://stackoverflow.com/a/73359981) and I couldn't get it to work. 

The docker image is 90mb after this change, whereas it was 23mb previously (3.6 times bigger). In the future, we could try building against musl (Alpine's libc) and move back to Alpine. But we had problems compiling against that in previous projects so it might not be straightforward.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,14 @@
-FROM alpine:latest
+FROM ubuntu:23.04
 
-RUN apk update
-RUN apk upgrade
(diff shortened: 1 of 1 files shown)


cfabianski:
I couldn’t come up with a better option for now

cfabianski (approved):

PR #934, opened 2023-04-24
ci: fix typo in linux build entrypoint

didroe:
## Description

Fix typo when checking for github user in linux build image entrypoint.

## Checklist


gabrielrinaldi:
@didroe do you know when this fix is going to be pushed to the docker image? I think it is breaking the GitHub Action.

didroe:
@gabrielrinaldi This is a fix for building the executables, images, etc. It doesn't effect the artifacts themselves. 

I think the problem with the action is because we changed from Alpine to Ubuntu for the base image, and the action entrypoint needs updating accordingly. I'm looking into it now.

PR #938, opened 2023-04-24
feat: add flag to disable version check

### LINKED ISSUE #892: New flag or env var to disable version check
Version check takes a little time as it must query github api in some circumstances its not very useful:
- integration testing of a known version
- ci when you are using docker with latest image
- you do not have external acess so the check will always fail

This would be a small optimization for these usecases.

elsapet:
## Description

Add flag to disable version check (for e.g. CI runs)

Closes #892

## Checklist

### DIFF
--- a/pkg/flag/general_flags.go
+++ b/pkg/flag/general_flags.go
@@ -29,25 +29,34 @@ var (
 		Usage:           "Load configuration from the specified path.",
 		DisableInConfig: true,
 	}
+	DisableVersionCheckFlag = Flag{
+		Name:       "disable-version-check",
+		ConfigName: "disable-version-check",
+		Value:      false,
+		Usage:      "Disable Bearer version checking",
+	}
 )
 
 type GeneralFlagGroup struct {
-	ConfigFile *Flag
-	APIKey     *Flag
-	Host       *Flag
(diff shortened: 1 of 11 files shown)


pkg/flag/general_flags.go
```diff
@@ -29,25 +29,34 @@ var (
 		Usage:           "Load configuration from the specified path.",
 		DisableInConfig: true,
 	}
+	DisableVersionCheckFlag = Flag{
+		Name:       "disable-version-check",
+		ConfigName: "disable-version-check",
+		Value:      false,
+		Usage:      "Disable Bearer version checking",
```
cfabianski, line 36: Do you wanna use the `Hide` attribute so that it doesn’t show up?

elsapet: Happy to hide it, but I'm curious the reason why?

gotbadger: @cfabianski see #892 for use cases and why this is a useful option.

elsapet: @cfabianski are you happy to leave this one visible?

gotbadger (approved):

PR #949, opened 2023-05-02
feat: add length_less_than filter

elsapet:
## Description

Context: current less than / greater than filters work on integer values only. 

Add length less than filter to support string values here. 

Relates to https://github.com/Bearer/bearer-rules/pull/34

## Checklist

### DIFF
--- a/new/detector/implementation/custom/filter.go
+++ b/new/detector/implementation/custom/filter.go
@@ -5,6 +5,7 @@ import (
 
 	"golang.org/x/exp/slices"
 
+	"github.com/bearer/bearer/new/detector/implementation/generic"
 	"github.com/bearer/bearer/new/detector/types"
 	"github.com/bearer/bearer/new/language/tree"
 	languagetypes "github.com/bearer/bearer/new/language/types"
@@ -46,7 +47,7 @@ func matchFilter(
 		)
 	}
 
-	return matchContentFilter(filter, node.Content()), nil, nil
+	return matchContentFilter(filter, evaluator, node), nil, nil
 }
 
 func matchAllFilters(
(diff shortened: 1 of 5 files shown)


elsapet:
Not sure if we want to add the other ones also (greater than, less than or equal to). Any thoughts?

new/detector/implementation/custom/filter.go
```diff
@@ -175,6 +175,12 @@ func matchContentFilter(filter settings.PatternFilter, content string) *bool {
 		return boolPointer(filter.Regex.MatchString(content))
 	}
 
+	if filter.LengthLessThan != nil {
+		if len(content) >= *filter.LengthLessThan {
```
didroe: I think it would be better to get the string detection to compare against. That way quotes won't be included, and we can support additional cases where variables are used.

In the case there is no string detection, we'd return `nil` - we use ternary logic for the number filters so "not a number" is always falsey, even when you wrap it in `not`.

new/detector/implementation/javascript/string/string.go
```diff
@@ -54,11 +53,15 @@ func concatenateChildren(node *tree.Node, evaluator types.Evaluator) ([]interfac
 			continue
 		}
 
-		childValue, childIsLiteral, err := getStringValue(child, evaluator)
+		childValue, childIsLiteral, err := generic.GetStringValue(child, evaluator)
 		if err != nil {
 			return nil, err
 		}
 
+		if childValue == "" {
```
didroe: This should prevent empty string literals `""` being considered a variable:
```suggestion
		if childValue == "" && !childIsLiteral {
```

new/detector/implementation/javascript/string/string.go
```diff
@@ -80,11 +83,15 @@ func handleTemplateString(node *tree.Node, evaluator types.Evaluator) ([]interfa
 		text += partText
 		return nil
 	}, func(child *tree.Node) error {
-		childValue, childIsLiteral, err := getStringValue(child.Child(1), evaluator)
+		childValue, childIsLiteral, err := generic.GetStringValue(child.Child(1), evaluator)
 		if err != nil {
 			return err
 		}
 
+		if childValue == "" {
```
didroe:
```suggestion
		if childValue == "" && !childIsLiteral {
```

didroe (approved):

PR #951, opened 2023-05-02
feat: ignore local 'insecure' URLs

### LINKED ISSUE #867: improve http insecure rules for js and ruby
## Description & Reproduction

By parsing battle test results we have found that we have some amount of false positive detections for using http instead of https.
The thing is most of these detections should be easy to ignore since they start either with http://localhost or http://127.0.0.1

## Expected Behavior

Ignore http insecure detections that start with localhost or 127.0.0.1

elsapet:
## Description

In the insecure URL detector, disregard URLs that start with `http://localhost` and `http://127.0.0.1` 
These are insecure URLs but we do not want to flag them because they are for local development

Closes #867 

## Checklist


cfabianski (commented):
Could we write a test for it?

cfabianski (approved):
Thank you

PR #953, opened 2023-05-02
ci: fix pathing issue with kpi scan

gotbadger:
## Description
Run the scans from the project dir so the file paths are correct.

## Checklist


elsapet (approved):
🎉

PR #954, opened 2023-05-02
fix: check if object is nil

vjerci:
## Description
This pr adds a fix for nil exception

## Checklist

### DIFF
--- a/new/detector/implementation/java/.snapshots/TestJavaObjects-object_class
+++ b/new/detector/implementation/java/.snapshots/TestJavaObjects-object_class
@@ -1,0 +1,57 @@
+- position: "1:1"
+  content: |-
+    public class User
+    {
+        public String name
+
+        public String LowercaseName()
+        {
+            return name.toLowerCase()
+        }
+    }
+  data:
+    properties:
+        - name: User
+          node: {}
+          object:
+            detectortype: object
+            matchnode: {}
+            data:
+                properties:
+                    - name: name
+                      node: {}
+                      object:
+                        detectortype: object
+                        matchnode: {}
+                        data: null
+                isvirtual: false
+    isvirtual: false
+- position: "1:1"
+  content: |-
+    public class User
+    {
+        public String name
+
+        public String LowercaseName()
+        {
+            return name.toLowerCase()
+        }
+    }
+  data:
+    properties:
+        - name: User
+          node: {}
+          object:
+            detectortype: object
+            matchnode: {}
+            data:
+                properties:
+                    - name: LowercaseName
+                      node: {}
+                      object:
+                        detectortype: object
+                        matchnode: {}
+                        data: null
+                isvirtual: false
+    isvirtual: false
+
--- a/new/detector/implementation/java/.snapshots/TestJavaObjects-object_class
+++ b/new/detector/implementation/java/.snapshots/TestJavaObjects-object_class
@@ -2,11 +2,11 @@
   content: |-
     public class User
     {
-        public String name
+        public String name;
 
         public String LowercaseName()
         {
-            return name.toLowerCase()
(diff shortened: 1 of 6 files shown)


cfabianski (commented):
Have a test file maybe?

vjerci:
> Have a test file maybe?

added it

new/detector/implementation/java/testdata/class.java
```diff
@@ -0,0 +1,9 @@
+public class User
+{
+    public String name
```
gotbadger:
```suggestion
    public String name;
```

new/detector/implementation/java/testdata/class.java
```diff
@@ -0,0 +1,9 @@
...
+
+    public String LowercaseName()
```
gotbadger, line 5:
```suggestion
    public String LowercaseName();
```

new/detector/implementation/java/testdata/class.java
```diff
@@ -0,0 +1,9 @@
...
+    {
+        return name.toLowerCase()
```
gotbadger:
```suggestion
        return name.toLowerCase();
```

new/detector/implementation/java/.snapshots/TestJavaObjects-object_class
```diff
@@ -0,0 +1,57 @@
...
+  data:
+    properties:
+        - name: User
+          node: {}
+          object:
+            detectortype: object
+            matchnode: {}
+            data:
+                properties:
+                    - name: LowercaseName
```
didroe, line 35: Not related to this change, but something I can see now you've added the tests:

It would be nice if these properties were both on the same object, rather than having two objects with a single property.

didroe (approved):

new/detector/implementation/java/.snapshots/TestJavaObjects-object_class
vjerci: fixed, ps. It feels weird to define an object name inside the property as a field called `name` feels like I'm naming property while I'm naming an object.
would be better if we could define object name once and `property` define their own names, or picks them up from their `Object` property.
```
	object := generictypes.Object{
		IsVirtual: false,
	}

	for _, property := range properties {
		object.Properties = append(object.Properties, generictypes.Property{
			Name:   className,
			Node:   node,
			Object: property,
		})
	}
```

would be better if it was something like this
```
	object := generictypes.Object{
		IsVirtual: false,
		Name: className,
	}

	for _, property := range properties {
		object.Properties = append(object.Properties, generictypes.Property{
			Node:   node,
			Object: property,
		})
	}
```

didroe: Once you have nested objects, there's not a distinction between object and property name. It's both depending on how you look at it. 

We could add a special case for the top level but I don't think it's worth it. You also have anonymous objects where there isn't a name. eg. `someCall({ key: value })`.

Looking at the code again, I think there's still something wrong. There should be a single property for the class at the top level:

```
{ User: { name, LowercaseName } }
```

But it's currently:

```
{ User: { name }, User: { LowercaseName } }
```

vjerci: yeah ok you're right. Could you write the code for me, I honestly don't understand how this property thing works.

As far as I can tell from the code it belongs to the same object with multiple properties, making me think the code is ok but suspect something is wrong with the way we export things. 
Maybe it's just some code gotcha.

https://github.com/Bearer/bearer/pull/954/files#diff-f2858e7bb9c4cf328b23d5a8c3543c6d51b3819021dddda306e52944026c870bR151

PR #959, opened 2023-05-03
fix: produce a single property for java classes

didroe:
## Description

Changes the Java object detector so we only produce a single property for the class. Previously a class like:

```java
    public class User
    {
        public String name;

        public String LowercaseName()
        {
            return name.toLowerCase();
        }
    }
```

would produce an object: `{ User: { name }, User: { LowercaseName } }`
when it should be: `{ User: { name, LowercaseName } }`.

We also remove the separate object detections for methods and fields as they're not needed.

## Checklist

### DIFF
--- a/new/detector/implementation/java/.snapshots/TestJavaObjects-object_class
+++ b/new/detector/implementation/java/.snapshots/TestJavaObjects-object_class
@@ -12,7 +12,7 @@
   data:
     properties:
         - name: User
-          node: {}
+          node: null
           object:
             detectortype: object
             matchnode: {}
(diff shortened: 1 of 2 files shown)


vjerci (commented):
Maybe I'm missing something.
However, I have doubts about the approach of having large tree sitter queries, I would prefer If we would implement framework changes that would allow us to merge nonvirtual objects and their properties.

didroe:
@vjerci I think it will be okay performance wise, but I can split it back into two queries if you prefer.

Regarding the detections themselves though, I don't think it makes sense to have an object detection for each method/field. They aren't objects in of themselves.

vjerci (approved):
ok, i see your point

PR #964, opened 2023-05-04
feat: add filename filter

didroe:
## Description

Add a new type of filter for filenames. The first use case will be a rule that matches Rails configuration, but ignoring test/development config files.

## Checklist


elsapet (approved):
Neat 😎

cfabianski (commented):
Love it! Would be useful for routes.rb as well (where we have password detection) WDYT?

PR #991, opened 2023-05-16
fix: handle flow correctly for sanitizer rules

didroe:
## Description

Moves the responsibility for sanitization rules to the evaluator. This allows us to terminate the match properly in the presence of flow/unified nodes.

Previously we wouldn't consider this case sanitized:

```ruby
x = params[:oops]
y = sanitize(x)

unsafe_call(y)
```

## Checklist


elsapet (approved):
Looks good 🙌 

For my interest, are there any related bearer-rules changes, or will these come later?

didroe:
@elsapet Yeah there are rule changes that I'll be making a PR for soon

PR #995, opened 2023-05-18
feat: official support for shared rules

didroe:
## Description

It's currently possible to put common patterns into separate rules, and then use these from another rule. This PR makes that more explicit/officially supported:
- Support for a new rule type `shared`
- Ability to list a number of other rules as `imports`
- Some basic validation of rules, including:
  - checking all rules referenced from a filter/sanitizer are either local or imported.
  - shared rules have a limited set of attributes, eg. no severity or remediation message.
- When `--only-rule` is used, automatically include any needed imported rules.

## Checklist


elsapet (approved):
👏

PR #997, opened 2023-05-19
fix: exclude doc folders and files from gitleaks detection

elsapet:
## Description

While doing the KPI scans on real-world repositories such as GitLab we found a lot of "secrets" that were not secrets but were examples in a markdown file in a docs folder.

This PR proposes to exclude any `.md` files within a `doc/s` folder from the gitleaks check. 

As a related note, further false positives were found on e.g. grafana project but these were within test data folders making them a bit trickier to isolate (i.e. not clear `.md` formats). 

## Checklist

### DIFF
--- a/pkg/detectors/gitleaks/gitlab_config.toml
+++ b/pkg/detectors/gitleaks/gitlab_config.toml
@@ -840,5 +840,6 @@ keywords = [
 description = "global allow lists"
 paths = [
     '''gitleaks.toml''',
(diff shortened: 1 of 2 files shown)


cfabianski:
It's always tricky to exclude and make such a strong decision on whether or not we should exclude.
I was looking at [this](https://gitlab.com/gitlab-org/security-products/analyzers/secrets/-/blob/master/gitleaks.toml) for instance and they don't seem to ignore those files.

elsapet:
Thanks @cfabianski - a good point. I'll close this.

PR #1009, opened 2023-05-24
fix: allow metadata to be set when incorrect

gotbadger:
## Description
In some CI contexts rev-parse can be unreliable give the option to specify env vars to use instead.

## Checklist

### DIFF
--- a/pkg/report/output/output.go
+++ b/pkg/report/output/output.go
@@ -170,11 +170,10 @@ func reportSecurity(
 		var meta *Meta
 		meta, err = getMeta(config)
 		if err != nil {
-			log.Debug().Msgf("couldn't get meta for repo %s", err)
-			config.Client.Error = pointer.String("Could not extract git metadata.")
-			meta = &Meta{
-				Target: config.Scan.Target,
-			}
+			errorMessage := fmt.Sprintf("Unable to calculate Metadata. %s", err)
+			log.Debug().Msgf(errorMessage)
+			config.Client.Error = &errorMessage
+			return
 		}
 
 		tmpDir, filename, err := createBearerGzipFileReport(config, meta, securityResults, dataflow)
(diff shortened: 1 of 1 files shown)


pkg/report/output/output.go
```diff
@@ -170,11 +170,10 @@ func reportSecurity(
 		var meta *Meta
 		meta, err = getMeta(config)
 		if err != nil {
-			log.Debug().Msgf("couldn't get meta for repo %s", err)
-			config.Client.Error = pointer.String("Could not extract git metadata.")
-			meta = &Meta{
-				Target: config.Scan.Target,
-			}
+			errorMessage := fmt.Sprintf("Unable to calulate Metadata. %s", err)
```
cfabianski:
```suggestion
			errorMessage := fmt.Sprintf("Unable to calculate Metadata. %s", err)
```

pkg/report/output/output.go
```diff
@@ -294,9 +341,9 @@ func getMeta(config settings.Config) (*Meta, error) {
 		FullName:           info.FullName,
 		URL:                info.Raw,
 		Target:             config.Scan.Target,
-		SHA:                strings.TrimSuffix(string(sha), "\n"),
-		CurrentBranch:      strings.TrimSuffix(string(currentBranch), "\n"),
-		DefaultBranch:      strings.TrimPrefix(strings.TrimSuffix(string(defaultBranch), "\n"), "origin/"),
+		SHA:                strings.TrimSuffix(*sha, "\n"),
```
cfabianski: Not a big thing but this is more for the cmd Output than for the os.Getenv so maybe we could locate the TrimSuffix where it makes more sense?

cfabianski (approved):

cfabianski (approved):

PR #1015, opened 2023-05-29
feat: add no-color flag

### LINKED ISSUE #917: Provide a `no-color` flag
We disable the color when we specify the an `output` file. 
Ideally we should also provide a way to disable the color by passing the option `no-color` (or `disable-color`?) so that we can use `bearer scan . > file.md` for instance

elsapet:
## Description

Add no-color flag so that we can disable colour output. Defaults to `false`

Example usage: 

```
  bearer scan . --no-color=true
```

Closes #917 

## Checklist

### DIFF
--- a/docs/_data/bearer_scan.yaml
+++ b/docs/_data/bearer_scan.yaml
@@ -51,6 +51,9 @@ options:
       default_value: '[]'
       usage: |
         Define regular expressions for better classification of private or unreachable domains e.g. --internal-domains=".*.my-company.com,private.sh"
+    - name: no-color
+      default_value: "false"
+      usage: Disable color in output
     - name: only-rule
       default_value: '[]'
       usage: |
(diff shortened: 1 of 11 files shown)


didroe (commented):
I think this would be better as a global option, rather than specific to the report. Although I appreciate we may only have the report generating colour at the moment.

If we can honor `NO_COLOR` (https://no-color.org/) that would be nice too

elsapet:
@didroe good shout on the `NO_COLOR` convention. I've tested this locally, and it looks like fatih color library already respects `NO_COLOR` env var https://pkg.go.dev/github.com/fatih/color@v1.15.0#NoColor - is that enough for us or shall we ensure it on Bearer's side too?

didroe:
@elsapet I think that's ok, if that's the only place we're outputting colour.

didroe (approved):

elsapet:
@didroe it's the only colour library we're using directly

PR #1019, opened 2023-05-30
fix: ruby element reference query

### LINKED ISSUE #1005: Multiple results returned for query error
## Description & Reproduction

Scan the following code

```ruby
def bug
  foo = [:a, :b, :c, :d, :e]
  logger.info(foo[0,2].join(":"))

  logger.info(current_user.email) # should be found
end
```

## Expected Behavior

The scan reports the findings (open redirect, sensitive data in log)

## Actual Behavior

No findings are reported

## Possible Fix

Running detectors report yields the following error: 

```
failed to process file for detector : ruby failed to detect in file example.rb: query returned more than one result
```

## Your Environment

- Output of 'bearer version':
```
bearer version: main
sha: 814937fa9c00701facfdcedc911615d1047cb3ea
```

elsapet:
## Description

Our element reference query was matching on things like `foo_array[0, 2]`. This was causing issue with the pattern matching returning multiple results. 
To address this, we add some anchors to the query so that we now match only on the simple hash access case e.g. `foo[:bar]`.

Closes #1005 

## Checklist

### DIFF
--- a/new/detector/composition/ruby/ruby.go
+++ b/new/detector/composition/ruby/ruby.go
@@ -192,10 +192,14 @@ func (composition *Composition) DetectFromFileWithTypes(file *file.FileInfo, det
 	var result []*detectortypes.Detection
 	for _, detectorType := range detectorTypes {
 		rule := composition.rules[detectorType]
+		sanitizerRuleID := ""
+		if rule != nil {
+			sanitizerRuleID = rule.SanitizerRuleID
+		}
 		detections, err := evaluator.Evaluate(
 			tree.RootNode(),
 			detectorType,
-			rule.SanitizerRuleID,
+			sanitizerRuleID,
 			settings.NESTED_SCOPE,
 			false,
 		)
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,5 +21,7 @@ jobs:
         uses: actions/setup-go@v4
         with:
           go-version: 1.19
-      - name: Run tests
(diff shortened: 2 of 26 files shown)


didroe (approved):
You could add a multiple index case here:
https://github.com/Bearer/bearer/blob/main/new/detector/implementation/ruby/testdata/object_projection.rb

new/detector/composition/ruby/ruby.go
```diff
@@ -192,10 +192,14 @@ func (composition *Composition) DetectFromFileWithTypes(file *file.FileInfo, det
...
+		sanitizerRuleID := ""
+		if rule != nil {
+			sanitizerRuleID = rule.SanitizerRuleID
+		}
 		detections, err := evaluator.Evaluate(
 			tree.RootNode(),
 			detectorType,
-			rule.SanitizerRuleID,
+			sanitizerRuleID,
 			settings.NESTED_SCOPE,
```
elsapet, line 203: I'll re-visit this tomorrow - @didroe I'm guessing a `nil` rule here is not a real-world possibility?

didroe: You should always have a rule as the `detectorTypes` were gathered from the rules in the first place

elsapet: ^ From discussion with @didroe : some built-in detectors e.g. `datatype` won't have a rule so we are okay to handle the `nil` case here (it's not just for the tests)

PR #1023, opened 2023-05-31
feat: add fingerprint to report and move skip rule to docs

elsapet:
## Description
With the introduction of the `--exclude-fingerprint` flag, we replace the `--skip-rule=rule_id` line in the security report with the finding's fingerprint `--exclude-fingerprint=0fd..._1`

Screenshot 2023-05-31 at 12 25 58

We move the more general `--skip-rule` copy to the documentation itself

Screenshot 2023-05-31 at 11 28 13

## Checklist


markmichon (approved):
Nice! 👍

PR #1024, opened 2023-05-31
fix: result scope following testing

didroe:
## Description

- Fixes `effectiveScope` by ensuring we only convert nested scope to result scope. Prior to this change, we were converting cursor scope into result scope.
- Run sanitizer rules in the default nested scope

## Checklist

### DIFF
--- a/new/detector/implementation/custom/filter.go
+++ b/new/detector/implementation/custom/filter.go
@@ -45,7 +45,7 @@ func matchFilter(
 
 	if filter.Detection != "" {
 		effectiveScope := filter.Scope
-		if scope == settings.RESULT_SCOPE {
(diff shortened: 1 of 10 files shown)


new/detector/implementation/custom/filter.go
```diff
@@ -45,7 +45,7 @@ func matchFilter(
 
 	if filter.Detection != "" {
 		effectiveScope := filter.Scope
-		if scope == settings.RESULT_SCOPE {
+		if filter.Scope == settings.NESTED_SCOPE && scope == settings.RESULT_SCOPE {
```
elsapet: Not sure if this is clearer 🤔 

```suggestion
		if effectiveScope == settings.NESTED_SCOPE && scope == settings.RESULT_SCOPE {
```

elsapet (approved):

PR #1034, opened 2023-06-05
docs: add redirect for removed ruby encryption rule

elsapet:
## Description
Since we have replaced the Ruby weak encryption rule by several separate encryption rules, we add a redirect for the old docs page

Relates to https://github.com/Bearer/bearer-rules/pull/83

## Checklist

### DIFF
--- a/docs/_redirects
+++ b/docs/_redirects
@@ -8,3 +8,4 @@
 /reference/rules/javascript_express_insecure_template_rendering /reference/rules/javascript_lang_raw_html_using_user_input
 /reference/rules/javascript_aws_lambda_os_command_injection /reference/rules/javascript_lang_os_command_injection
 /reference/rules/javascript_aws_lambda_query_injection /reference/rules/javascript_third_parties_dynamodb_query_injection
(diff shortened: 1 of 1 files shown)


docs/_redirects
```diff
@@ -8,3 +8,5 @@
 /reference/rules/javascript_express_insecure_template_rendering /reference/rules/javascript_lang_raw_html_using_user_input
 /reference/rules/javascript_aws_lambda_os_command_injection /reference/rules/javascript_lang_os_command_injection
 /reference/rules/javascript_aws_lambda_query_injection /reference/rules/javascript_third_parties_dynamodb_query_injection
+
+/reference/rules/ruby_lang_weak_encryption /reference/rules
```
elsapet, line 11: @markmichon wondering if it's better (less confusing) to redirect to the rules index page as there's no single replacement here - or do we prefer to select one of the encryption rules?

markmichon: Normally it's not great practice, but for this use case it should be fine.

markmichon (approved):

PR #1075, opened 2023-06-19
perf: various performance enhancements

### LINKED ISSUE #1065: Some rules cause scan to hang for a few seconds.
## Description & Reproduction

Scan hangs when processing some javascript projects

Checkout https://github.com/Bearer/NodeGoat

hangs:
```
~/bearer scan ./app/assets/vendor/chart/raphael-min.js --force ---debug --only-rule=javascript_third_parties_algolia
```
ok: 
```
~/bearer scan ./app/assets/vendor/chart/raphael-min.js --force --debug --only-rule=javascript_third_parties_airbrake
```

## Expected Behavior

Scan does not hang

- Output of 'bearer version':
```
bearer version: 1.11.0
sha: ba142fa855a36da17e94e080df09861b8c660516
```

didroe:
## Description

Improves performance by:
- Caching detections at the evaluation level rather than just the node level. This avoids repeated walking of the same part of the tree
- When a file times out, try to gracefully stop the scan in the worker rather than having to kill the process. This avoids reloading the rules in this case.
- Don't create a tree sitter query for patterns that are just a single variable. These match any node anyway.
- Sort filters in order of likely complexity before evaluation. This helps to throw away invalid matches before running the more expensive filters.

Benchmark:

|Project|Before|After|Speedup|
|-|-|-|-|
|Rocket Chat|2min 11sec|1min 36sec|1.4x|
|Juice Shop|1min 27sec|1min 13sec|1.2x|

_(numbers are from the file progress bar, not entire program execution time)_

Also:
- Fixes an issue where a debug profile would not be written if the worker was interrupted
- Adds `--log-level` option and adds some trace logging for rule evaluation.
- Adds rule timing to debug output when `--debug-profile` is set
- Changes log messages to pure text rather than JSON. This is much better when there are newlines in the output.
- Fixes a panic in the typescript global data types detector.

## Related
- Close #1065

## Checklist

### DIFF
--- a/e2e/flags/.snapshots/TestInitCommand
+++ b/e2e/flags/.snapshots/TestInitCommand
@@ -13,12 +13,12 @@ rule:
 scan:
     context: ""
     data_subject_mapping: ""
-    debug: false
     disable-domain-resolution: true
     domain-resolution-timeout: 3s
     external-rule-dir: []
     force: false
     internal-domains: []
+    log-level: info
     parallel: 0
     quiet: false
     scanner:
--- a/e2e/flags/.snapshots/TestMetadataFlags-help-scan
+++ b/e2e/flags/.snapshots/TestMetadataFlags-help-scan
@@ -24,12 +24,13 @@ Rule Flags
 Scan Flags
       --context string                       Expand context of schema classification e.g., --context=health, to include data types particular to health
       --data-subject-mapping string          Override default data subject mapping by providing a path to a custom mapping JSON file
-      --debug                                Enable debug logs
+      --debug                                Enable debug logs. Equivalent to --log-level=debug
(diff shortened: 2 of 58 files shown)


e2e/flags/.snapshots/TestInitCommand
```diff
@@ -13,12 +13,12 @@ rule:
 scan:
     context: ""
     data_subject_mapping: ""
-    debug: false
```
cfabianski, line 16: Can we keep that one (deprecated) and make it pass the right value to the new `log-level` instead?

didroe: I think it should only be omitted from generating the default config, but still supported for existing files.

cfabianski (approved):

PR #1080, opened 2023-06-20
feat: support filters when referencing rules

didroe:
## Description

Adds support for specifying filters when using a rule reference (`detection`) filter. 

This will allow us to write shared rules that are then further filtered in rules that use them. eg. we can write a shared rule to match all Java variable definitions, and then filter on the type in rules that use it.

## Checklist


cfabianski (approved):
very nice!

PR #1083, opened 2023-06-21
fix: support variables in js imports

didroe:
## Description

Adds support for variables in place of package names in Javascript imports. eg. `import x from $<VARIABLE>`.

This is currently not supported because we always generate a dummy identifier for a variable (eg. `curioVar1`) but the grammar requires a string (eg. `"curioVar1"`).

To cope with this, we introduce a mechanism for fixing up patterns. Firstly the pattern is parsed with the default variable behaviour, then we look for `ERROR` nodes and attempt to fixup the variables. Afterwards, we attempt a parse again.

Support for Ruby classes is also added.

Following this change, node types should no longer need to be specified in rules.

## Checklist

### DIFF
--- a/e2e/flags/report_flags_test.go
+++ b/e2e/flags/report_flags_test.go
@@ -14,6 +14,7 @@ func newScanTest(name string, arguments []string) testhelper.TestCase {
 	arguments = append([]string{
 		"scan",
 		"--disable-version-check",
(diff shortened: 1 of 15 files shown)


cfabianski (approved):
It was a matter of time 😅

PR #1101, opened 2023-06-28
chore: include file in errors in dataflow

cfabianski:
## Description

Add the files for which we had an error while processing it.

e.g.

```json
  ...
  "errors": [
    {
      "type": "file_error",
      "filename": "/frontend/src/assets/private/three.js",
      "error": "file processing time exceeded"
    }
  ]
  ...
```

Also comment out the blamer since we don't make use of it at the moment.

## Checklist


pkg/commands/process/worker/worker.go
```diff
@@ -56,13 +55,13 @@ func (worker *Worker) Setup(config config.Config) error {
 }
 
 func (worker *Worker) Scan(ctx context.Context, scanRequest work.ProcessRequest) error {
-	blamer := blamer.New(scanRequest.Dir, scanRequest.BlameRevisionsFilePath, scanRequest.PreviousCommitSHA)
+	// blamer := blamer.New(scanRequest.Dir, scanRequest.BlameRevisionsFilePath, scanRequest.PreviousCommitSHA)
```
didroe: I think we should remove the code rather than comment it out

pkg/detectors/openapi/openapi.go
```diff
@@ -44,7 +44,7 @@ func (detector *detector) ProcessFile(file *file.FileInfo, dir *file.Path, repor
 
 	fileType, err := getFileType(file)
 	if err != nil {
-		return false, err
+		return false, nil
```
didroe, line 49: Is this intentional? If so, can we log the error (as debug)?

didroe (approved):

pkg/detectors/openapi/openapi.go
cfabianski: Good idea!

PR #1107, opened 2023-06-30
feat: add more helpful rule validation

### LINKED ISSUE #429: Add validator for YAML rules
We should validate rule format and rule patterns, in order to provide adequate feedback to users who are writing their own custom rules. Validation errors could include YAML line references and links to rule writing documentation. 

Such validation could be iterative, for example, and start simply by validating the required YAML attributes are present. 

We should also consider validating the external rule patterns themselves during load time, to avoid throwing obscure error messages during run time. See related issue here (partially addressed) https://github.com/Bearer/bearer/issues/545.

gotbadger:
## Description
Currently when creating a rule if the rule is invlaid the feedback is very poor. Here is an example with a very old version of `airbrake.yml:`:
```
Loading rules
Error: error loading external rules from /Users/phil/bearer-rule2: failed to unmarshal rule airbrake.yml: yaml: unmarshal errors:
  line 52: cannot unmarshal !!str `local` into settings.RuleDefinitionTrigger
  line 54: cannot unmarshal !!map into string
```
With this change we get the following:
```
Loading rules

Failed to load airbrake.yml
Validating against https://raw.githubusercontent.com/Bearer/bearer-rules/main/scripts/rule_schema.json

airbrake.yml validation issues found:
- (root): Must validate "else" as "if" was not valid
- metadata: documentation_url is required
- metadata: Additional property dsr_id is not allowed
- trigger: Invalid type. Expected: object, given: string
- severity: Invalid type. Expected: string, given: object

Error: external rules rule file was invalid
```

## Related
closes #429

## Checklist

### DIFF
--- a/pkg/commands/process/settings/ruleValidator.go
+++ b/pkg/commands/process/settings/ruleValidator.go
@@ -1,0 +1,68 @@
+package settings
+
+import (
+	"bytes"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/ghodss/yaml"
+	"github.com/xeipuuv/gojsonschema"
+)
+
+const SCHEMA_URL = "https://raw.githubusercontent.com/Bearer/bearer-rules/main/scripts/rule_schema.json"
+
+func ValidateRule(entry []byte, filename string) string {
+	validationStr := &strings.Builder{}
+	validationStr.WriteString(fmt.Sprintf("Failed to load %s\nValidating against %s\n\n", filename, SCHEMA_URL))
+	schema, err := loadSchema(SCHEMA_URL)
+	if err != nil {
+		validationStr.WriteString("Could not load schema to validate")
+		return validationStr.String()
+	}
+
+	jsonData, err := yaml.YAMLToJSON(entry)
+	if err != nil {
+		validationStr.WriteString("File fdormat is invalid")
+		return validationStr.String()
+	}
+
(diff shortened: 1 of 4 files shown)


pkg/commands/process/settings/ruleValidator.go
```diff
@@ -0,0 +1,68 @@
...
+	"bytes"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/ghodss/yaml"
+	"github.com/xeipuuv/gojsonschema"
+)
+
+const SCHEMA_URL = "https://raw.githubusercontent.com/Bearer/bearer-rules/main/scripts/rule_schema.json"
```
cfabianski, line 13: hum relying on main feels a bit dangerous. What do you think?

gotbadger: Its the easiest way to get the schema otherwise I think we will need to download the sourcecode from the latest release and extract it that way. 

Bear in mind that the schema is only loaded if a rule does not unmarshal for some reason so you only end up using the schema if a rule is totally broken. Since this is the case I think mains fine to use.

gotbadger: Other option might be to store the schema on the docs site or something 🤔 

Let me have a think about this one.

cfabianski (approved):

PR #1127, opened 2023-07-18
docs: add doc on how to exclude findings

gmontard:
## Description

Add documentation entry on how to exclude findings.


elsapet (approved):
💯

PR #1134, opened 2023-07-19
fix: respect gitignore syntax in skip-path

elsapet:
## Description

We should support full gitignore syntax with the `skip-path` flag. 

Our current library did not support ignoring subdirectories, so the following skip-path pattern would not skip subdirectories e.g. `public/static/temp/1.js`

```
  skip-path: [public]
```

I propose a different library that does not have the limitations of `filepath.Match`. Note: this may carry some performance considerations. 

## Checklist


cfabianski (approved):

elsapet:
No apparent difference in performance between this branch and `main` (tested with and without `skip-path` set)

PR #1137, opened 2023-07-21
feat: improve code extract to be consistent

### LINKED ISSUE #54: Inaccurate Ruby logger code snippets and line numbers
## Description & Reproduction

The console report for Ruby lang and Rails logger findings is confusing. 
The line number given corresponds to the sensitive data and not the inclusion of this sensitive data within the logger call. The code snippet (and so, presumably, the parent line numbers) correspond to the logger call itself, but since the line number is not within the parent line number, the snippet shows without any content or highlight.

### Repro steps

1. Run a scan on the following code snippet (in a file): 

```ruby
class User < ApplicationRecord
  def get_user(email)
    user_params = {
      primary_email: UserEmail.new(email: email, primary: true)
    }

    Rails.logger.warn(
      "New User (user_id: #{user.id}) Params: #{user_params}"
    )

    logger.error(
      "New User (user_id: #{user.id}) Params: #{user_params}"
    )

    user
  end
end
```

2. Examine the report given

## Expected Behavior

Code snippet should include relevant pieces of code

## Actual Behavior

Line number is not accurate
Code snippet is empty

Screenshot 2023-05-15 at 13 17 33

## Your Environment

- Output of 'bearer version':

```
bearer version: 1.6.1
sha: 743e8d8a246db1225670f0e34766cfc139cb3d5a
```

Note also seen on 
```
bearer version: main
sha: bbf6cc415923f13742da6d0399f26d92190ec47f```

### PR DESCRIPTION
**Pull request by cfabianski**
## Description

Code Extract is not consistent between CLI and Cloud because the CLI reads the lines from the file while the Cloud gets the info from the Source and Sink which may not be the entire line.
The new JSON field `code_extract` corrects that and makes it consistent

Also:
 - Refactored the way we generate the omitted line and logic which were a little bit too complicated.
 - Makes it easier to configure the buffer for line

Tested with HTML, JSON and Table outputs

Still ToDo:

- [ ] Requires a Cloud updates to start consuming this new field cc @gotbadger 
- [ ] Update bearer-rules snapshots 🙈 

image
image

## Related
- Closes https://github.com/Bearer/bearer-rules/issues/54

## Checklist

### DIFF
--- a/new/detector/composition/java/.snapshots/TestScope--scope.yml
+++ b/new/detector/composition/java/.snapshots/TestScope--scope.yml
@@ -28,7 +28,6 @@ high:
       snippet: scopeCursor(request.getParameter("oops"))
       fingerprint: bdbeee20feb34c6881d975716e2fe09f_0
       old_fingerprint: bdbeee20feb34c6881d975716e2fe09f_0
-      code_extract: ""
     - rule:
         cwe_ids:
             - "42"
@@ -58,7 +57,6 @@ high:
       snippet: scopeNested(request.getParameter("oops"))
       fingerprint: bdbeee20feb34c6881d975716e2fe09f_1
       old_fingerprint: bdbeee20feb34c6881d975716e2fe09f_1
-      code_extract: ""
     - rule:
         cwe_ids:
             - "42"
@@ -88,7 +86,6 @@ high:
       snippet: 'scopeNested(x ? request.getParameter("oops") : y)'
       fingerprint: bdbeee20feb34c6881d975716e2fe09f_2
       old_fingerprint: bdbeee20feb34c6881d975716e2fe09f_2
-      code_extract: ""
     - rule:
         cwe_ids:
             - "42"
@@ -118,7 +115,6 @@ high:
       snippet: 'scopeNested(request.getParameter("oops") ? x : y)'
       fingerprint: bdbeee20feb34c6881d975716e2fe09f_3
       old_fingerprint: bdbeee20feb34c6881d975716e2fe09f_3
-      code_extract: ""
     - rule:
         cwe_ids:
             - "42"
@@ -148,7 +144,6 @@ high:
       snippet: scopeResult(request.getParameter("oops"))
       fingerprint: bdbeee20feb34c6881d975716e2fe09f_4
       old_fingerprint: bdbeee20feb34c6881d975716e2fe09f_4
-      code_extract: ""
     - rule:
         cwe_ids:
             - "42"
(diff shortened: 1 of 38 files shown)

### CODE REVIEW

File: pkg/report/output/html/security.tmpl [hunk 1]
```diff
@@ -23,7 +23,7 @@
           </div>
 
           <p class="filename">Filename: {{.Filename}}:{{.LineNumber}}</p>
-          <div class="term-container">{{. | displayExtract}}</div>
+          <pre class="term-container">{{. | displayExtract}}</pre>
```
gotbadger: Dont think this needs to be changed?

cfabianski: It's not required but I noticed it during my test that it respects better the spaces and it looks better.
I'm joining the before/after for comparison
WDYT?

**with Pre**
image

**with Div**
image

gotbadger: The font is being overidden by the browser to be monospaced when using the pre tag. You dont need to use a pre tag to do this. @Ombe-b designed it like this.

new/detector/composition/java/.snapshots/TestScope--scope.yml
```diff
@@ -28,6 +28,7 @@ high:
       snippet: scopeCursor(request.getParameter("oops"))
       fingerprint: bdbeee20feb34c6881d975716e2fe09f_0
       old_fingerprint: bdbeee20feb34c6881d975716e2fe09f_0
+      code_extract: ""
```
gotbadger: seems like there might be an issue here?

cfabianski: no it's expected

pkg/report/output/html/styles.css
```diff
@@ -170,7 +170,7 @@ table {
 .finding .description code.language-java {
   display: flex;
   padding: 8px;
-  font-family: Source Sans Pro, ui-sans-serif, system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, "Noto Sans", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
+  font-family: monospace, Sans Pro, ui-sans-serif, system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, "Noto Sans", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
```
gotbadger:
```suggestion
  font-family: ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace;
```

pkg/report/output/html/styles.css
```diff
@@ -227,6 +227,7 @@ span.warning{
 }
 
 .term-container {
+  font-family: monospace;
```
gotbadger:
```suggestion
  font-family: ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace;
```

gotbadger (approved):

PR #1145, opened 2023-07-25
feat: add flag to ignore report failures

### LINKED ISSUE #1119: Allow failures without exiting with an error code
Today when the CLI run it exits with a fail status code (-1) when a finding is found, this is a good default behavior.
But, it's not always enough, especially when starting to implement such a solution into a CI/CD workflow, you don't necessarily want to fail it but only want to be able to see the report. 

For that reason, it would be important to add a new flag that disables failed status code whenever a finding is reported, for e.g:
```bearer scan . --allow_failure```

For reference, GitLab CI allow such behavior (cf [doc](https://docs.gitlab.com/ee/ci/yaml/#allow_failure)) but not GitHub Action, so it's probably better to bake such solution natively in Bearer CLI.

_PS: This need has been reported by a user._

### PR DESCRIPTION
**Pull request by gotbadger**
## Description
Adds a new flag that when used will not set the failure exit code dispite any errors from the scan. The tool will still fail on any other unrelated issues like misconfiguration etc.

## Related
- Close #1119

## Checklist

### DIFF
--- a/docs/_data/bearer_scan.yaml
+++ b/docs/_data/bearer_scan.yaml
@@ -1,103 +1,107 @@
-name: ' scan'
+name: " scan"
 synopsis: Scan a directory or file
-usage: ' scan [flags] <path>'
+usage: " scan [flags] <path>"
 options:
-    - name: api-key
-      usage: Use your Bearer API Key to send the report to Bearer.
-    - name: config-file
-      default_value: bearer.yml
-      usage: Load configuration from the specified path.
-    - name: context
-      usage: |
-        Expand context of schema classification e.g., --context=health, to include data types particular to health
-    - name: data-subject-mapping
-      usage: |
-        Override default data subject mapping by providing a path to a custom mapping JSON file
-    - name: debug
-      default_value: "false"
-      usage: Enable debug logs. Equivalent to --log-level=debug
-    - name: debug-profile
-      default_value: "false"
-      usage: Generate profiling data for debugging
-    - name: disable-default-rules
-      default_value: "false"
-      usage: Disables all default and built-in rules.
-    - name: disable-domain-resolution
-      default_value: "true"
-      usage: |
-        Do not attempt to resolve detected domains during classification
-    - name: disable-version-check
-      default_value: "false"
-      usage: Disable Bearer version checking
-    - name: domain-resolution-timeout
-      default_value: 3s
-      usage: |
-        Set timeout when attempting to resolve detected domains during classification, e.g. --domain-resolution-timeout=3s
-    - name: exclude-fingerprint
-      default_value: '[]'
-      usage: |
-        Specify the comma-separated fingerprints of the findings you would like to exclude from the report.
-    - name: external-rule-dir
-      default_value: '[]'
-      usage: |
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: docs/_data/bearer_scan.yaml [hunk 1]
```diff
@@ -2,6 +2,10 @@ name: ' scan'
 synopsis: Scan a directory or file
 usage: ' scan [flags] <path>'
 options:
+    - name: allow-failure
+      default_value: "false"
+      usage: |
+        Forces the exit code to be 0 (success) dispite any findings from the scan
```
elsapet:
```suggestion
        Forces the exit code to be 0 (success) despite any findings from the scan
```

pkg/flag/scan_flags.go
```diff
@@ -108,6 +108,12 @@ var (
 		Value:      0,
 		Usage:      "Specify the amount of parallelism to use during the scan",
 	}
+	AllowFailureFlag = Flag{
+		Name:       "allow-failure",
+		ConfigName: "scan.allow-failure",
+		Value:      false,
+		Usage:      "Forces the exit code to be 0 (success) dispite any findings from the scan",
```
elsapet:
```suggestion
		Usage:      "Forces the exit code to be 0 (success) despite any findings from the scan",
```

elsapet (approved):

docs/_data/bearer_scan.yaml
```diff
@@ -37,6 +37,10 @@ options:
       default_value: '[]'
       usage: |
         Specify the comma-separated fingerprints of the findings you would like to exclude from the report.
+    - name: exit-code
+      default_value: "-1"
+      usage: |
+        Forces the exit code to be 0 (success) despite any findings from the scan
```
elsapet: A suggestion to re-word this a bit, since now we can give any exit code we want

```suggestion
        Use this to force a given exit code for the scan command. Set this to 0 (success) to always return a success exit code despite any findings from the scan. 
```

elsapet (approved):

docs/_data/bearer_scan.yaml
```diff
@@ -1,103 +1,107 @@
...
+    usage: |
+      Set timeout when attempting to resolve detected domains during classification, e.g. --domain-resolution-timeout=3s
+  - name: exclude-fingerprint
+    default_value: "[]"
+    usage: |
+      Specify the comma-separated fingerprints of the findings you would like to exclude from the report.
+  - name: exit-code
+    default_value: "-1"
+    usage: |
+      Use this to force a given exit code for the scan command. Set this to 0 (success) to always return a success exit code despite any findings from the scan.
```
elsapet: Seeing the others, maybe this is more consistent. Sorry for the back and forth 🙈 
```suggestion
      Force a given exit code for the scan command. Set this to 0 (success) to always return a success exit code despite any findings from the scan.
```

elsapet (approved):

docs/_data/bearer_scan.yaml
cfabianski: No pb. Agreed!

elsapet (approved):
🙏 🙏 🙏 🙏 🙏

cfabianski (approved):

PR #1149, opened 2023-07-28
chore: update detector type when classification

cfabianski:
## Description

Detector Type was always set to Ruby

## Checklist

### DIFF
--- a/new/detector/implementation/generic/datatype/datatype.go
+++ b/new/detector/implementation/generic/datatype/datatype.go
@@ -30,17 +30,7 @@ type datatypeDetector struct {
 	classifier *classificationschema.Classifier
 }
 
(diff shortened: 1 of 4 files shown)


new/detector/implementation/generic/datatype/datatype.go
```diff
@@ -25,13 +25,25 @@ type Property struct {
 }
 
 type datatypeDetector struct {
+	detectorType detectors.Type
 	types.DetectorBase
 	classifier *classificationschema.Classifier
 }
 
-func New(lang languagetypes.Language, classifier *classificationschema.Classifier) (types.Detector, error) {
+func New(language string, lang languagetypes.Language, classifier *classificationschema.Classifier) (types.Detector, error) {
```
didroe: Should we just pass in the detector type rather than converting from language?

didroe (approved):

Issue #1144, opened 2023-07-25
Diff scan


didroe:
To avoid doing unnecessary work, support a new `--diff-base-branch` option which takes a git base branch name and does the following:
- Find changed or added files
- Checkout and scan base commit 
- Checkout and scan head commit
- Report findings from the head commit that aren't present in the base commit

The exact method of matching findings needs some investigation. Fingerprints may not be sufficient

PR #1158, opened 2023-07-31
feat: base branch diff

didroe:
## Description

Adds support for only reporting findings that have been introduced on a branch, relative to a base branch. This is only available for the security report, and not when using the SaaS API. 

The base branch is specified using the `DIFF_BASE_BRANCH` environment variable. The base branch will be fetched if it's not currently available.

There's no indication in the security report JSON/YML as it would require changing the format.

Documentation will be updated in a further PR once manual testing on Github/Gitlab is completed.

## Related

- #1144

## Checklist


cfabianski (approved):

joevin-slq-docto:
Hello @didroe,
Can you explain me a bit more how this works ?

I've made a first experiment with `export DIFF_BASE_BRANCH=master` on a branch which was 2 commits behind master and bearer stayed on for a long, long time on `Scanning target .`. But it finally worked.

Inside the GitHub action in the other hand, it failed :
```
Loading rules
Scanning target .
Error: scan error: error fetching base branch: error fetching branch master: authentication required
scan error: error fetching base branch: error fetching branch master: authentication required
Error: Process completed with exit code 1.
```

It seems to be a pure analysis of the two branches and then a comparison of the results. I understand that it helps keep alerts relevant, but performance-wise, I doubt it's much better than a full analysis (I will make tests).

I also have got concerns regarding the way of diffing two branches by directly using git.
At doctolib, in business hours it's not uncommon to have one commit each 5 minutes, and pull requests are not necessarily rebased from master before each github action. On the other hand, if we try to identify the differences between a branch's initial commit and the branch's current commit, the results will be polluted by the various master rebase commits.

didroe:
Hi @joevin-slq-docto,

The way it works is that it looks for what's changed between the base branch and the current/head commit, and only scans files that changed. Then the findings from the current commit are reported, if they didn't also appear in the base commit. Roughly speaking (and ignoring git time!), the scan should be faster if you changed < 50% of the files in the repo.

In terms of the git behaviour, it checks out the entire base branch at it's current commit. There are a couple of issues with the current approach:
- In terms of checking out files, the expectation is that most of the file data (git objects) will already be present so it shouldn't be too heavy to fetch/checkout the entire base commit (rather than just the files we want to scan). Perhaps that isn't the case in practice though - that could explain the pause on `Scanning target .`.
- We should be checking out the base commit of the branch you're working on (ie. the base branch at the commit you branched off), rather than the base branch's current commit. The current behaviour is incorrect here.

Regarding the authentication issue under CI: we are using a Go git library rather than the git executable in the environment. We made this decision as we've seen multiple bugs in the past due to different git versions. I guess the library doesn't read the auth info from the git config. I'll take a look into this.

Some questions for you that will help me to improve the feature:
- With your repo, do you think it would be a heavy operation to fetch and checkout all files at the base commit? (given the initial state/checkout of your repo during CI runs).
- Are you using Github's `checkout` action? How do you set up authentication for Git during CI? 
- Regarding your last paragraph, I think I've addressed the first sentence above. But in your last sentence (which will be the behaviour once I've fixed it), I don't understand what you mean by "the results will be polluted by the various master rebase commits". Can you explain further please?

joevin-slq-docto:
Hello @didroe,

Execution time of the git step is highly variable: sometimes it's a few seconds, sometimes it takes up to 9 minutes (after a git prune and on a branch created yesterday that is already 122 commits behind master).
We can expect this time to be munch more on the CI.

As discussed on slack, feel free to directly gave me git commands to put after actions/checkout@v3 on a github action.

For questions :

1. Via actions/checkout@v3, not a lot (~50s). But after, potentially yes : I know that a `git checkout my_branch` takes at least 2 minutes on the CI.
2. Yes, re-using git binary (and associated settings) of the runner seems to be sufficient to execute git commands.
3. I will try to explain more the sentense "the results will be polluted by the various master rebase commits".
Imagine this : 
```
   master
     | fff01
     | fff02     new-branch
     |---------------| aaa01 - Init commit of my new feature
     | fff03         |
     |               | aaa02 - Fixing thins on my new feature
     |               | aaa03 - Merge branch 'master' into 'new-branch'
     |---------------| aaa04 - My feature is fine, ready to be merged
```
If the scan is executed on files differences between `aaa01` and `aaa04`.
Even if new-branch only contains modifications on README.md.
The scan will be applied on all modifications on README.md and all of those introduced by the merge commit (and include fff03 differences).

Nb: I also noticed that bearer diff scan broke the husky git hook.

PR #1163, opened 2023-08-01
fix: repository lookup errors and diff git file scanning

didroe:
## Description

Makes the following fixes to changes introduced in https://github.com/Bearer/bearer/pull/1158 :
- Fix error trying to lookup git repo when scanning a single non-git file
- Allow diff scans to scan files/folders within the repo, rather than only the entire repo

## Checklist


cfabianski (commented):
Any tests to help see the issue? 😁

didroe:
@cfabianski I've added a test that would catch the original issue. We're lacking tests generally in this area unfortunately, but it's not that straightforward to add them.

cfabianski (approved):

PR #1169, opened 2023-08-02
docs: add diff scan

didroe:
## Description

Adds documentation for diff scanning, and simplifies reviewdog instructions for Github.

## Related

- #1144

## Checklist


cfabianski:
Happy with it. 
@gmontard I’ll let you have a look and approve or suggest changes :-)

cfabianski:
> Let's mention that diff scan makes the scan way faster - in a callout (it might seem obvious but let's make it very very clear :))
> Wondering if we need to explain that the integration with reviewdog doesn't change if we are using diff scan (diff scan just makes the scan faster 🤔)

Side Note
> Instead of "Set up CI/CD" let's have "Integrate with CI/CD"

gotbadger (approved):
@didroe please change the base branch to main

PR #1185, opened 2023-08-14
docs: switch to free and open terminology

gmontard:
Bearer is using the Elastic license, which implicate is free and open. Reflecting the right terminology in the read me.


gmontard:
Close https://github.com/Bearer/bearer/issues/1183

elsapet (approved):

PR #1198, opened 2023-08-22
ci: add trivy to scan docker image

cfabianski:
## Description

Add Trivy for scanning docker image

## Checklist


gotbadger:
maybe its just worth having this on canary instead since any problems would be revealed after it was already released?

gotbadger (approved):

PR #1202, opened 2023-08-22
refactor: extract report data object and formatters

### LINKED ISSUE #1201: Refactor report output process
We often want to include data in one report, but not another. This is currently quite tricky to do and so we end up including a lot of "extra" data in one report (e.g. JSON report) simply because it is needed in another (e.g. SaaS report). 

We can refactor this process to make this a bit easier. 

Proposed refactoring: 

1. Introduce some Report object that can hold (optional) data like Detectors, Dataflow, Security Findings, Findings Grouped By Severity
2. Add a formatter that takes this Report object and a format type, and produces a report output. This could be as simple as a map of formatting type to rendering function

### Related
- #1194
- #1195

elsapet:
## Description

Refactoring to further separate report data from report format

Main points: 
* Introduce a ReportData object that holds any data needed for the various report formats
* Extract formatters for each report type, and move bulk of the formatting work out of run.go

Minor points:
* Rename security result to security finding

Closes #1201 

## Checklist


pkg/report/output/saas/types/types.go
```diff
@@ -23,11 +23,16 @@ type Meta struct {
...
+	DataTypes  []dataflowtypes.Datatype  `json:"data_types" yaml:"data_types"`
+	Components []dataflowtypes.Component `json:"components" yaml:"components"`
+	Errors     []dataflowtypes.Error     `json:"errors" yaml:"errors"`
+	Files      []string                  `json:"files" yaml:"files"`
 	// Dependencies []dataflowtypes.Dependency    `json:"dependencies" yaml:"dependencies"`
 }
+
+type SaasFinding struct {
+	Finding        *securitytypes.Finding
+	SomeExtraField string `json:"some_extra_field" yaml:"some_extra_field"`
```
elsapet, line 37: To be removed - proof of concept to understand how we would include field in SaaS Findings but not Security Findings

pkg/report/output/dataflow/formatter.go
```diff
@@ -0,0 +1,37 @@
...
+		Config:     config,
+	}
+}
+
+func (f Formatter) Format(format string) (output *string, err error) {
+	switch format {
+	case flag.FormatEmpty:
+		output, err = outputhandler.ReportJSON(f.ReportData.Dataflow)
+	case flag.FormatJSON:
+		output, err = outputhandler.ReportJSON(f.ReportData.Dataflow)
```
cfabianski:
```suggestion
	case flag.FormatEmpty, flag.FormatJSON:
		output, err = outputhandler.ReportJSON(f.ReportData.Dataflow)
```

pkg/report/output/detectors/formatter.go
```diff
@@ -0,0 +1,37 @@
...
+		Config:     config,
+	}
+}
+
+func (f Formatter) Format(format string) (output *string, err error) {
+	switch format {
+	case flag.FormatEmpty:
+		output, err = outputhandler.ReportJSON(f.ReportData.Detectors)
+	case flag.FormatJSON:
+		output, err = outputhandler.ReportJSON(f.ReportData.Detectors)
```
cfabianski:
```suggestion
	case flag.FormatEmpty, flag.FormatJSON:
		output, err = outputhandler.ReportJSON(f.ReportData.Detectors)
```

pkg/report/output/privacy/formatter.go
```diff
@@ -0,0 +1,54 @@
...
+func NewFormatter(reportData *outputtypes.ReportData, config settings.Config) *Formatter {
+	return &Formatter{
+		ReportData: reportData,
+		Config:     config,
+	}
+}
+
+func (f Formatter) Format(format string) (output *string, err error) {
+	switch format {
+	case flag.FormatEmpty:
```
cfabianski, line 27: It feels like we should default to `json` and add a new `flag.FormatCSV` option. Not being able to get this `csv` other than not specifying a type feels like a smell to me

pkg/report/output/saas/formatter.go
```diff
@@ -0,0 +1,37 @@
...
+		Config:     config,
+	}
+}
+
+func (f Formatter) Format(format string) (output *string, err error) {
+	switch format {
+	case flag.FormatEmpty:
+		output, err = outputhandler.ReportJSON(f.ReportData.SaasReport)
+	case flag.FormatJSON:
+		output, err = outputhandler.ReportJSON(f.ReportData.SaasReport)
```
cfabianski:
```suggestion
	case flag.FormatEmpty, flag.FormatJSON:
		output, err = outputhandler.ReportJSON(f.ReportData.SaasReport)
```

cfabianski (commented):
Feels like a massive improvement already! 👍

pkg/report/output/privacy/formatter.go
elsapet: I think the default format for privacy report should remain CSV (otherwise it's quite a big change in functionality), but you're right, we could add a flag for this and have something like:

```
case flag.FormatEmpty, flag.FormatCSV:
```

elsapet: Added this in a separate PR here https://github.com/Bearer/bearer/pull/1208 - so that this PR can remain refactoring only (no functionality changes)

cfabianski (approved):

pkg/report/output/detectors/formatter.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func (f Formatter) Format(format string) (output *string, err error) {
+	switch format {
+	case flag.FormatEmpty, flag.FormatJSON:
+		output, err = outputhandler.ReportJSON(f.ReportData.Detectors)
+	case flag.FormatYAML:
+		output, err = outputhandler.ReportYAML(f.ReportData.Detectors)
+	default:
+		err = fmt.Errorf(`--report flag "%s" is not supported`, f.Config.Report.Report)
```
didroe: You could move the error to the caller and return a `nil` output here. That would avoid duplicating the error message in each formatter.

pkg/report/output/detectors/formatter.go
```diff
@@ -0,0 +1,35 @@
...
+	return &Formatter{
+		ReportData: reportData,
+		Config:     config,
+	}
+}
+
+func (f Formatter) Format(format string) (output *string, err error) {
+	switch format {
+	case flag.FormatEmpty, flag.FormatJSON:
+		output, err = outputhandler.ReportJSON(f.ReportData.Detectors)
```
didroe: You could also do:

```suggestion
		return outputhandler.ReportJSON(f.ReportData.Detectors)
```

pkg/report/output/output.go
```diff
@@ -22,99 +24,94 @@ import (
...
+
+	if config.Report.Report == flag.ReportDataFlow {
+		err = GetDataflow(data, report, config, false)
+		return data, err
+	}
+
+	// add dataflow to data for internal use
+	if err = GetDataflow(data, report, config, true); err != nil {
+		return data, err
+	}
```
didroe: We could use a single call here:
```suggestion
	if err = GetDataflow(data, report, config, config.Report.Report != flag.ReportDataFlow); err != nil {
		return data, err
	}
```

pkg/report/output/output.go
```diff
@@ -22,99 +24,94 @@ import (
...
-) (*types.Output[any], error) {
-	var err error
-	var output types.GenericOutput
+) (*types.ReportData, error) {
+	sendToCloud := false
+
+	data := &types.ReportData{}
+	// add detectors
+	err := detectors.AddReportData(data, report, config)
+	if config.Report.Report == flag.ReportDetectors {
```
didroe: I think we'll miss errors for non-detectors reports:
```suggestion
	if config.Report.Report == flag.ReportDetectors || err != nil {
```

didroe (approved):
Nice work! ❤️

PR #1206, opened 2023-08-24
perf: remove tree sitter walking

### LINKED ISSUE #1172: Tree sitter node caching
During a scan we do a lot of walking nodes to find children with particular states we are interested in for each rule. If we cache these findings in an internal representation it should be possible to speed up the overall scan time. 

This should lead to less timeout and even memory issues in workers.

didroe:
## Description

closes #1172 

## Checklist


.envrc.example
```diff
@@ -6,4 +6,7 @@ export REDIS_INIT="true"
 export GOOGLE_MAX_ATTEMPT="5"
 export BEARER_EXECUTABLE_PATH="./bearer"
 export GITHUB_WORKSPACE="/path/to/bearer/project"
-export SCAN_DIR=/Users/username/OWASP
\ No newline at end of file
+export SCAN_DIR=/Users/username/OWASP
+export BEARER_DISABLE_DEFAULT_RULES=true
+export BEARER_EXTERNAL_RULE_DIR=../bearer-rules/rules
+export BEARER_FORCE=true
```
cfabianski, line 12: Do we need to do `unset ${!BEARER_*}` when running the tests maybe? (in `run_tests.sh` I would have thought)

cfabianski (approved):

PR #1212, opened 2023-08-25
fix: improve output from ignore commands

gotbadger:
## Description
Improve the output from ignore commands

- format json so file is more friendly 
- beautify display of ignored entries
- show new ignored entry on add

## Checklist

### DIFF
--- a/pkg/util/ignore/ignore.go
+++ b/pkg/util/ignore/ignore.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"path/filepath"
 	"time"
+
+	"github.com/fatih/color"
 )
 
 type IgnoredFingerprint struct {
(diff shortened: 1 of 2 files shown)


pkg/util/ignore/ignore.go
```diff
@@ -59,7 +89,7 @@ func AddToIgnoreFile(fingerprintsToIgnore map[string]IgnoredFingerprint, force b
 		existingIgnoredFingerprints[key] = value
 	}
 
-	data, err := json.Marshal(existingIgnoredFingerprints)
+	data, err := json.MarshalIndent(existingIgnoredFingerprints, "", "  ")
```
elsapet, line 92: Nice!

elsapet (approved):
🙌

PR #1222, opened 2023-08-28
fix: bearer-ignore-flag not loaded for scan

elsapet:
## Description

We weren't including ignore flags during scan, so bearer.ignore was not being loaded.
Also ensure we look to the right place if scan target is a path

## Checklist


elsapet:
For completeness, we'd probably want to add this flag to bearer actions too (as we have done for `--config-file`). What do you think @gotbadger

gotbadger (approved):

gotbadger:
maybe move `bearer-ignore-file` to general flags since its used by more than just ignore commands?

pkg/commands/ignore.go
```diff
@@ -263,13 +263,14 @@ $ bearer ignore migrate`,
 			if err != nil {
 				return fmt.Errorf("flag error: %s", err)
 			}
+
 			configFilePath := viper.GetString(flag.ConfigFileFlag.ConfigName)
```
elsapet, line 267: Curious here: is the following equivalent? 

```suggestion
			configFilePath := viper.GetString(options.GeneralOptions.ConfigFile)
```

We use `flag.ConfigFileFlag.ConfigName` in the commands/scan.go but perhaps this is for performance or other reasons. Any light to be shed @didroe or @cfabianski? 🙏

didroe: In this case it's equivalent (without the `viper.GetString`). But in the scan command, we support reading the options from the config file. So we have to read it's location before building the options.

elsapet: Ahh, thanks for the explanation. So, if we wanted to support reading options from the config file in this case too, we should leave it as the following?

```golang
configFilePath := viper.GetString(flag.ConfigFileFlag.ConfigName)
```

didroe: Yeah. Basically we'd need to do [this](https://github.com/Bearer/bearer/blob/911c5c161774f639f1d34e7dc25a7b4596b8df82/pkg/commands/scan.go#L79-L88) before calling `<FLAGS>.ToOptions`. And therefore you'd need to grab the filename manually first.

PR #1223, opened 2023-08-28
feat: read from config file for ignore migrate

elsapet:
## Description

Read options from config file (if present) when migrating ignores. See discussion for further context: https://github.com/Bearer/bearer/pull/1222#pullrequestreview-1598398125

## Checklist

### DIFF
--- a/pkg/commands/ignore.go
+++ b/pkg/commands/ignore.go
@@ -6,9 +6,7 @@ import (
 	"os"
 
 	"github.com/bearer/bearer/pkg/flag"
-	"github.com/bearer/bearer/pkg/util/file"
 	"github.com/bearer/bearer/pkg/util/ignore"
-	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


pkg/commands/ignore.go
```diff
@@ -259,13 +261,30 @@ $ bearer ignore migrate`,
...
+			if err := readConfig(configPath); err != nil {
+				if err := readConfig(defaultConfigPath); err != nil {
+					loadFileMessage = "Couldn't find any config file"
+				} else {
+					loadFileMessage = fmt.Sprintf("Loading default config file %s", defaultConfigPath)
+				}
+			} else {
+				loadFileMessage = fmt.Sprintf("Loading config file %s", configPath)
+			}
+			log.Debug().Msgf(loadFileMessage)
```
didroe: Can we put this in a function and use it in both places?

didroe (approved):

didroe (approved):

PR #1244, opened 2023-09-07
docs: add reference to PHP support coming next

gmontard:
## Description
Add reference to PHP support coming next:
- Top of GH README
- In the language support doc page 

image

image

### DIFF
--- a/docs/reference/supported-languages.njk
+++ b/docs/reference/supported-languages.njk
@@ -124,7 +124,7 @@ When you scan a codebase, Bearer will automatically select the appropriate langu
         </td>
         <td>
           {{ data.status }}<br/>
-          <small><i>{{ data.comment }}</i> </small>
+          <span class="italic text-xs">{{ data.comment }}</span>
         </td>
       </tr>
     {% endfor %}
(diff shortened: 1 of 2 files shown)


gotbadger (approved):

docs/reference/supported-languages.njk
```diff
@@ -122,7 +123,8 @@ When you scan a codebase, Bearer will automatically select the appropriate langu
           {% endif %}
         </td>
         <td>
-          {{ data.status }}
+          {{ data.status }}<br/>
+          <small><i>{{ data.comment }}</i> </small>
```
gotbadger:
```suggestion
          <span class="italic text-sm">{{ data.comment }}</span>
```
Try and stick to using tailwind stuff as the browser styles from these tags are not consistant:
https://tailwindcss.com/docs/font-size

if text-sm is too big you can use `text-[12px]` class for one offs

gmontard: Used xs and looking great, thanks!

gotbadger (approved):

PR #1248, opened 2023-09-08
feat: add ignore pull action

### LINKED ISSUE #1246: Add bearer ignore pull action
Once https://github.com/Bearer/bearer/issues/1204 is in use, the local `bearer.ignore` file may become out-of-sync with the Cloud over time, if for example ignored findings are later re-opened in the Cloud. 
We therefore introduce a new ignore action that will allow us to overwrite the contents of the `bearer.ignore` file with the state from the Cloud. 

Note: This overwrite will be total - this means that any new entries in the `bearer.ignore` file added since the latest scan will be lost. Because of its destructive nature, the action **must include a warning message** with user confirmation to approve the overwrite. 

### Suggested interface

```bash
⇒ bearer ignore pull /path/to/project --api-key=123

Warning: this action will overwrite any new additions to your bearer.ignore file added since your last Cloud scan. Continue? [Y/n]
```

elsapet:
## Description

* Add `bearer ignore pull` action
* Hardcode log level to `error` for all ignore actions

Closes #1246

## Checklist

### DIFF
--- a/pkg/commands/ignore.go
+++ b/pkg/commands/ignore.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/bearer/bearer/pkg/flag"
+	"github.com/bearer/bearer/pkg/report/output/saas"
 	"github.com/bearer/bearer/pkg/util/ignore"
 	ignoretypes "github.com/bearer/bearer/pkg/util/ignore/types"
 	"github.com/spf13/cobra"
@@ -24,6 +25,7 @@ Available Commands:
     add              Add an ignored fingerprint
     show             Show an ignored fingerprint
     remove           Remove an ignored fingerprint
+    pull             Pull ignored fingerprints from Cloud
     migrate          Migrate ignored fingerprints
 
 Examples:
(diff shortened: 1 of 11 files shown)


pkg/commands/ignore.go
```diff
@@ -53,9 +60,12 @@ Examples:
 		newIgnoreShowCommand(),
 		newIgnoreAddCommand(),
 		newIgnoreRemoveCommand(),
+		newIgnorePullCommand(),
 		newIgnoreMigrateCommand(),
 	)
 
+	zerolog.SetGlobalLevel(zerolog.ErrorLevel)
```
elsapet, line 67: Currently `--debug` and `--log-level` flags are Scan flags, so they are not available in this action. 
Do we want to move them to General flags, so we can use them here?

gotbadger: Think this would make a lot of sense

didroe (approved):

pkg/commands/ignore.go
elsapet: ^ https://github.com/Bearer/bearer/pull/1257

PR #1249, opened 2023-09-08
feat: improve version check and rule download cache

### LINKED ISSUE #1088: Use known compatible rules
We currently always download the latest rules. If a rule change is made that requires newer functionality in the CLI (and isn't backwards compatible), we end up with the latest rule set failing to scan in older CLI versions.

We need a way to allow older CLI versions to continue running with rules they are compatible with.

### LINKED ISSUE #1135: Use location other than GitHub to download the rulesets
On GitLab, you are more likely to hit a Rate Limit if you try to download the ruleset for every PR since you are unauthenticated.
We could use on own servers to host and download the ruleset (as an option) which would remove that limitation.

## Proposal 

Perform an API call to our server and download the assets from there (they would be hosted on S3)

## Considerations

We would like to keep the information regarding the number of downloads of each ruleset

gotbadger:
## Description

- Version checking not using GitHub API
- Fallback to GitHub API
- Improve downloaded rule caching
- Timeouts for API requests
- Fixes issue where tests were downloading rules

## Related

closes #1135
closes #1088 (option3)

## Checklist


docs/_data/bearer_ignore_pull.yaml
```diff
@@ -0,0 +1,40 @@
...
+      usage: Set log level (error, info, debug, trace)
+    - name: no-color
+      default_value: "false"
+      usage: Disable color in output
+example: |-
+    # Pull ignored fingerprints from the Cloud (requires API key)
+    $ bearer ignore pull /path/to/your_project --api-key=XXXXX
+see_also:
+    - ' ignore - Manage ignored fingerprints'
+aliases:
```
cfabianski, line 40: Why is this file included in this PR?

cfabianski (approved):

docs/_data/bearer_ignore_pull.yaml
gotbadger: its missing in main apparenly and was generated when I ran the script

PR #1257, opened 2023-09-12
feat: move log level and debug flags to General

elsapet:
## Description

Debug and Log Level flags were Scan flags but we would like to use them in other contexts too (e.g. bearer ignore actions) and so we move them to General flags. 

The related docs are also updated. 

## Checklist


gotbadger (approved):

didroe (approved):
We'll need to call `output.Setup` in all the commands that want to use these flags.

PR #1260, opened 2023-09-13
fix: bearer ignore path when target is a file

elsapet:
## Description

Rule repo test failures tell us that more sophistication is needed when constructing ignore path. 

Updated flow is the following: 

If the user has passed a value to `--ignore-file`, we try this path first. If no file is found, we append the target directory path and try again. If still, no file is found, we return an error. 

If no value is passed to `--ignore-file`, we try `bearer.ignore` first. If no file is found, we append the target directory path and try again. If still no file is found, we assume it cannot be found. Depending on the action (e.g. `bearer ignore add`), a new file is then created in this location. 

## Checklist


cfabianski (approved):
LGTM

PR #1269, opened 2023-09-15
fix: default ignore file value

elsapet:
## Description

Updates to the recent changes around `--ignore file` flag: 

* Add default value back for flag
* Return actual path to ignore file and use this in console messages
* Fix migrate and pull commands (which weren't using `ignore-file` value)
* In copy, prefer more general "ignore file" (similar to e.g. "config file") to "`bearer.ignore` file"
* Update docs accordingly

## Checklist


cfabianski (approved):
Looks great

PR #1271, opened 2023-09-15
fix: show warning when unsupported languages are detected

### LINKED ISSUE #1229: Improve security scan output when only partial language is supported to not be deceptive to users.
Its currenly a bit deceptive when a repo only has partial support for a language say javascript but there are PHP also in the project.

When running a scan look for package information from languages we dont support and warn if we see it. Link to language support page of docs.

elsapet:
## Description

Show warning message when scan detects unsupported language(s)

Closes #1229

```bash
Security Report

=====================================

Rules:
https://docs.bearer.com/reference/rules [v0.18.3]

Language    Default Rules  Custom Rules  Files
JavaScript  72             0             3
Python      0              0             35

Warning: Bearer currently offers only partial support for Python. 
For more information, see https://docs.bearer.com/reference/supported-languages

```

## Checklist


internal/detectors/dependencies/dependencies.go
```diff
@@ -40,6 +40,54 @@ func (detector *detector) AcceptDir(dir *file.Path) (bool, error) {
 	return true, nil
 }
 
+func DetectorsForLanguage(language string) []string {
+	switch language {
+	case "C#":
```
elsapet: Wondering if we prefer to add language information to the dependencies directly, and avoid this case statement and maintenance thereof. Any thoughts?

elsapet: ^ After discussion, we will add language information to the dependencies, and ensure that this matches the gocloc result (i.e. update `"Javascript"` to `"JavaScript"`)

elsapet:
Snapshot changes a bit annoying here - we snapshot the object itself not the JSON / YML so the `omitempty` is not applied, and we have a lot of blank entries added.

cfabianski (approved):
Thanks!

PR #1285, opened 2023-09-21
fix: reporting of unsupported languages

didroe:
## Description

Fixes the output of unsupported languages. We were showing incorrect information with lots of duplicates.

Also sorts the list of languages for the warning message to give consistent output.

## Checklist


elsapet (approved):
🙏

PR #1290, opened 2023-09-22
fix: avoid compiling privacy report twice

elsapet:
## Description

We were compiling the privacy report again, when building the CSV string. We should check to see if it's already present in the report data before we do this

## Checklist

### DIFF
--- a/internal/report/output/privacy/privacy.go
+++ b/internal/report/output/privacy/privacy.go
@@ -68,12 +68,6 @@ const PLACEHOLDER_VALUE = "Unknown"
 func BuildCsvString(reportData *outputtypes.ReportData, config settings.Config) (*strings.Builder, error) {
 	csvStr := &strings.Builder{}
 	csvStr.WriteString("\nSubject,Data Types,Detection Count,Critical Risk Finding,High Risk Finding,Medium Risk Finding,Low Risk Finding,Rules Passed\n")
-	if reportData.PrivacyReport == nil {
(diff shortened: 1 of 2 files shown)


internal/report/output/privacy/privacy.go
```diff
@@ -68,9 +68,11 @@ const PLACEHOLDER_VALUE = "Unknown"
 func BuildCsvString(reportData *outputtypes.ReportData, config settings.Config) (*strings.Builder, error) {
 	csvStr := &strings.Builder{}
 	csvStr.WriteString("\nSubject,Data Types,Detection Count,Critical Risk Finding,High Risk Finding,Medium Risk Finding,Low Risk Finding,Rules Passed\n")
-	err := AddReportData(reportData, config)
```
cfabianski, line 71: In which circumstances can it be nil?
I'm wondering if we shouldn't simply remove that `AddReportData` call from here. WDYT?

elsapet: No current circumstances. I added it as a precaution, because `BuildCsvString` is a public method and there's no guarantee that the `reportData` we pass to it will contain a privacy report. Edit: that said, we do not check in `AddReportData` that e.g. `reportData.Dataflow` is present so perhaps this caution is inconsistent. I can remove it.

cfabianski: Let's remove it 👍

cfabianski (approved):

PR #1304, opened 2023-10-04
feat: allow diffscans to be sent to bearer cloud

gotbadger:
## Description
- We want to start processing the output of diffscans in cloud so allow them to be sent
- Impoved error messaging a bit

## Checklist

### DIFF
--- a/internal/commands/process/settings/settings.go
+++ b/internal/commands/process/settings/settings.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"time"
 
+	"golang.org/x/exp/slices"
 	"gopkg.in/yaml.v3"
 
 	"github.com/bearer/bearer/api"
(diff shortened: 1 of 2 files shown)


internal/commands/process/settings/settings.go
```diff
@@ -364,13 +364,9 @@ func FromOptions(opts flag.Options, versionMeta *version_check.VersionMeta) (Con
 	}
 
 	if config.Scan.DiffBaseBranch != "" {
-		if config.Report.Report != flag.ReportSecurity {
+		if config.Report.Report != flag.ReportSecurity && config.Report.Report != flag.ReportSaaS {
```
cfabianski:
```suggestion
		if !slices.Contains([]string{flag.ReportSecurity, flag.ReportSaaS}, config.Report.Report) {
```

❓

cfabianski (commented):
Is it worth having a test for it?
Do we need to pass a flag in the meta to know it's coming from a diff scan at all?

gotbadger:
@cfabianski was already sent in the meta and as you can see its just passing the value of the setting:
https://github.com/Bearer/bearer/blob/main/internal/report/output/saas/saas.go#L201

All this PR is really doing is allowing it to be sent now when it wasn't before.

cfabianski (approved):

PR #1305, opened 2023-10-04
fix(php): unanchor class name in new object expressions

didroe:
## Description

Allow `new Foo` to match `new Foo()` by not anchoring the `Foo` node within the object creation expression.

## Checklist


cfabianski (approved):
Do you want to write a test for it or are they any tests in the bearer-rules repo that will fail if that changes?

cfabianski (approved):

Issue #1302, opened 2023-10-04
Golang Alpha Support


cfabianski:
We've decided to support Golang. This issue will track overall development.

- analyzer bindings
- strings
- objects

PR #1314, opened 2023-10-10
feat(golang): add beta support

cfabianski:
## Description

Add Beta support for Golang

## Related

#1302 

## Checklist


internal/util/progressbar/progressbar.go
```diff
@@ -18,8 +18,9 @@ func GetProgressBar(filesLength int, config settings.Config, display_type string
 		progressbar.OptionOnCompletion(func() {
 			output.ErrorWriter().Write([]byte("\n")) //nolint:all,errcheck
 		}),
-		progressbar.OptionShowIts(),
-		progressbar.OptionSetItsString(display_type),
+		// progressbar.OptionShowIts(),
+		// progressbar.OptionSetItsString(display_type),
+		// progressbar.OptionClearOnFinish(),
```
didroe: Is it intentional to include this change?

didroe (approved):
I'm not sure about the language template. I like the idea, but I worry it will be hard to keep it up-to-date - my thinking is that something not used by the product will likely slowly diverge from the real languages, even with the best effort to apply changes to it

cfabianski:
> I'm not sure about the language template. I like the idea, but I worry it will be hard to keep it up-to-date - my thinking is that something not used by the product will likely slowly diverge from the real languages, even with the best effort to apply changes to it

Will delete the template after I'm done with Python. I agree.

internal/util/progressbar/progressbar.go
cfabianski: I will do it as a separate PR

PR #1321, opened 2023-10-11
feat: remove useless it/s information in progressbar

cfabianski:
## Description

Remove useless iteration / sec in the progress bar.

## Checklist

### DIFF
--- a/internal/util/progressbar/progressbar.go
+++ b/internal/util/progressbar/progressbar.go
@@ -6,7 +6,7 @@ import (
 	"github.com/schollz/progressbar/v3"
 )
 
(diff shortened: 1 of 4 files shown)


internal/util/progressbar/progressbar.go
```diff
@@ -18,8 +18,6 @@ func GetProgressBar(filesLength int, config settings.Config, display_type string
 		progressbar.OptionOnCompletion(func() {
 			output.ErrorWriter().Write([]byte("\n")) //nolint:all,errcheck
 		}),
-		progressbar.OptionShowIts(),
-		progressbar.OptionSetItsString(display_type),
```
gotbadger, line 22: display_type passed in is no longer used so can be cleaned up.

PR #1324, opened 2023-10-12
chore: update tree sitter

### LINKED ISSUE #1348: Upgrade tree sitter and associated parsers
Update tree sitter and all the parsers we bundle. We can also drop some of the vendored parsers that are now bundled/updated in the Go library.

We've fallen behind on versions and often see differences from the Tree Sitter Playground when developing rules/languages.

didroe:
## Description

- Update the Go Tree Sitter library
- Remove vendored toml and protobuf parsers as these are now in the Tree Sitter library
- Update other vendored parsers, including a rebase of our HTML parser fork 
- Update PHP string handling to account for grammar changes
- Update some snapshots due to incidental changes - including improvements to the reported column number for HTML script tag content.

There seems to be an issue with the newer Tree Sitter and matching wildcard nodes `(_)`. Queries don't always match unless such nodes have a field name or variable specified. To work around this, we now include the field name in the query.

Also changes the end-to-end tests to ignore any `BEARER_` variables defined in the environment. This allows the tests to be run locally with an `.envrc` that sets such variables.

## Related
- Close #1348 

## Checklist


cfabianski (approved):
Well done!

PR #1327, opened 2023-10-13
fix: cloud uploads on diff scans

elsapet:
## Description

During a diff scan, we first run a full scan and then the diff scan. We were uploading the first, full, scan to the Cloud which was causing unexpected findings to appear in the PR diff. 

## Checklist

### DIFF
--- a/internal/report/output/output.go
+++ b/internal/report/output/output.go
@@ -9,6 +9,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/hhatto/gocloc"
 	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
 
 	"github.com/bearer/bearer/internal/commands/process/settings"
 	"github.com/bearer/bearer/internal/flag"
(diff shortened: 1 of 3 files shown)


gotbadger (approved):

internal/report/output/output.go
```diff
@@ -74,15 +71,19 @@ func GetData(
...
-	}
-
 	return data, err
 }
 
+func UploadReportToCloud(report *types.ReportData, config settings.Config) {
+	if slices.Contains([]string{flag.ReportSecurity, flag.ReportSaaS}, config.Report.Report) {
+		if config.Client != nil && config.Client.Error == nil {
+			// send SaaS report to Cloud
+			report.SendToCloud = true
```
didroe: Is this flag used for anything now?

didroe (approved):

internal/report/output/output.go
elsapet: Good spot - it's not!

PR #1340, opened 2023-10-18
fix(classification): improve php datatype detections

didroe:
## Description

- Add missing cases to PHP object detector
- Strip `$` from names
- Strip `get` or `set` prefix from names

## Checklist


cfabianski (approved):
👏

PR #1344, opened 2023-10-20
perf: avoid allocations with `(*regexp.Regexp).MatchString`

Juneezee:
## Description

We should use `(*regexp.Regexp).MatchString` instead of `(*regexp.Regexp).Match([]byte(...))` when matching string to avoid unnecessary `[]byte` conversions and reduce allocations.

Example benchmark:

```go
func BenchmarkMatch(b *testing.B) {
	for i := 0; i < b.N; i++ {
		if match := looksLikeUrl.Match([]byte("https://www.example.com")); !match {
			b.Fail()
		}
	}
}

func BenchmarkMatchString(b *testing.B) {
	for i := 0; i < b.N; i++ {
		if match := looksLikeUrl.MatchString("https://www.example.com"); !match {
			b.Fail()
		}
	}
}
```

Result:

```
goos: linux
goarch: amd64
pkg: github.com/bearer/bearer/internal/parser/interfaces/paths cpu: AMD Ryzen 7 PRO 4750U with Radeon Graphics
BenchmarkMatch-16          	 1788562	       812.1 ns/op	      24 B/op	       1 allocs/op
BenchmarkMatchString-16    	 2106370	       517.8 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/bearer/bearer/internal/parser/interfaces/paths	3.628s
```

## Checklist

### DIFF
--- a/internal/detectors/custom/custom.go
+++ b/internal/detectors/custom/custom.go
@@ -387,7 +387,7 @@ func filterCaptures(params []config.Param, captures []parser.Captures) (filtered
 
 				paramContent := capture[param.BuildFullName()].Content()
 
-				if !regex.Match([]byte(paramContent)) {
+				if !regex.MatchString(paramContent) {
 					shouldIgnore = true
 					break
 				}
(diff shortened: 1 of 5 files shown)


cfabianski:
Thanks @Juneezee will wait for the CI and merge ❤️

PR #1350, opened 2023-10-23
ci: add version comparison workflow

didroe:
## Description

Adds a new manual workflow for comparing a test version of the CLI/rules against a baseline version. The security report is generated as JSON for both versions, and then a diff of the reports is outputted. If there are differences then the workflow fails.

The base/test CLI/rules versions can be specified as a git tag, branch name, or commmit SHA. Leaving a ref blank uses the latest commit on the default branch.

Also synchronizes the KPI repository list with those in the battle test repo.

## Checklist


kpi_scan/kpi_repo_list.json5
```diff
@@ -0,0 +1,36 @@
+{
```
gotbadger, line 1: json5?

cfabianski (approved):

PR #1376, opened 2023-11-06
chore(cli): set version for root

cfabianski:
## Description

Setting the version will automatically enable `bearer --version`.
I found it particularly annoying/confusing in some occasions to not have it available.

image

---

https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#version-flag

> Cobra adds a top-level '--version' flag if the Version field is set on the root command. Running an application with the '--version' flag will print the version to stdout using the version template. The template can be customized using the cmd.SetVersionTemplate(s string) function.

## Checklist

### DIFF
--- a/e2e/flags/.snapshots/TestMetadataFlags-version
+++ b/e2e/flags/.snapshots/TestMetadataFlags-version
@@ -1,5 +1,4 @@
 
 --
-bearer version: dev
-sha: devSHA
+bearer version dev, build devSHA
 
--- a/internal/commands/app.go
+++ b/internal/commands/app.go
@@ -66,7 +66,7 @@ Learn More:
 	For more examples, tutorials, and to learn more about the project
 	visit https://docs.bearer.com
 `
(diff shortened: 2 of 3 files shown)


gotbadger (dismissed):
This would basically mean we have 2 versions of the version command. I dont think we should do this. Maybe we could talk about it.

didroe (approved):
❤️

cfabianski:
> This would basically mean we have 2 versions of the version command. I dont think we should do this. Maybe we could talk about it.

Well I agree it's not ideal to have 2 ways to show the version. IMO the `bearer version` isn't good / standard (I often got caught by it myself). It only exists because we had not implemented it correctly in the first place. 
The only reason I hadn't removed the previous way as part of this PR was to be backward compatible but I would have removed it. Do you think I should do it?

gotbadger:
Bearer command structure is modelled after tools like docker which use this patten. So its not right to say its implmented incorrectly. It just is implemented like that :)

PR #1404, opened 2023-11-17
feat: improve testing for rules

### LINKED ISSUE #1405: Improve testing for rules
It would be useful to rely on comments instead of snapshots for the rules.

Having a comment where we expect the rule to trigger for example. That way, it would be easier to write the tests and it would improve the debugging (we could display the flow and the rule triggered along with the decision process) if something doesn't match our expectation.

https://github.com/Bearer/bearer/compare/spike/improve-testing-for-rules?expand=1

cfabianski:
## Description

Closes https://github.com/Bearer/bearer/issues/1405

The idea is to allow comments to flag that a finding is expected.
That way we can improve the way we tests 

See https://github.com/Bearer/bearer-rules/pull/180

ToDo:

- [ ] Should we use a feature flag to turn this feature on as it might have some perf implication?

## Checklist


didroe (approved):
Nice!

PR #1443, opened 2023-12-18
chore: update viper

### LINKED ISSUE #1439: Improve flags testing coverage
As part of the viper update https://github.com/Bearer/bearer/pull/1431 we took the opportunity to resolve a FIXME

Though, it had to be reverted (despite some manual testing) https://github.com/Bearer/bearer/pull/1437 since following this release, the GitHub Action would not work (which is using Env variables)

We should increase our testing coverage on that area

cfabianski:
## Description

Closes #1439

Updates Viper

Also:
- [x] Make the binding public so that it can be tested
- [x] Improve test coverage for env variables (particularly repository flags)
- [x] Remove cycle dependency by splitting flag types.

In the end, the `AutomaticEnv` and the `SetEnvPrefix` wasn't the way to go because the values in `EnvironmentVariables` shouldn't used the prefixed version.
The current code is the final version which I believe is the simplest version.

## Checklist


didroe (approved):

cfabianski:
This PR would be useful if we wanted to improve it https://github.com/spf13/viper/pull/1719

PR #1464, opened 2024-01-12
feat: add more scan context for gitlab

gotbadger:
## Description
Add more scan context for gitlab, use IID for pull request.

## Checklist

### DIFF
--- a/internal/flag/repository_flags.go
+++ b/internal/flag/repository_flags.go
@@ -118,8 +118,8 @@ var (
 		Value:      "",
 		Usage:      "Used when fetching branch level ignores for a PR/MR",
 		EnvironmentVariables: []string{
-			"PR_NUMBER",           // github
-			"CI_MERGE_REQUEST_ID", //gitlab
+			"PR_NUMBER",            // github
+			"CI_MERGE_REQUEST_IID", //gitlab
 		},
 		DisableInConfig: true,
 		Hide:            true,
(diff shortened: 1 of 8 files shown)


internal/flag/repository_flags.go
```diff
@@ -118,8 +118,8 @@ var (
 		Value:      "",
 		Usage:      "Used when fetching branch level ignores for a PR/MR",
 		EnvironmentVariables: []string{
-			"PR_NUMBER",           // github
-			"CI_MERGE_REQUEST_ID", //gitlab
```
cfabianski, line 122: I would have expected some tests for this 🙈

internal/report/output/saas/saas.go
```diff
@@ -213,5 +213,6 @@ func getMeta(
 		BearerVersion:      build.Version,
 		FoundLanguages:     reportData.FoundLanguages,
 		GitlabPipelineId:   os.Getenv("CI_PIPELINE_ID"),
+		GitlabJobId:        os.Getenv("CI_JOB_ID"),
```
cfabianski, line 216: Do you not want to add a test for this report (while you are at it)?

internal/flag/repository_flags.go
cfabianski: ahah, the test is failing so it works :D

cfabianski (approved):
When tests are green

cfabianski (approved):
Thank you 😘

PR #1469, opened 2024-01-19
feat: remove dynamic severity

### LINKED ISSUE #1468: Remove Dynamic Severity Level
## Reasoning

- Dynamic severity level is a nightmare
    - Users don’t understand it
    - We have edge-cases where we create confusion (diff scan without the context etc.)
    

What we’re trying to do here is “threat-modeling” for the AppSec team, mostly at the application level, considering rules directly tied to sensitive data are very limited.

## Proposal

- Remove dynamic severity level from CLI and Cloud, for findings
- Create a “business risk score” using the App sensitive data detect  that we can show on the Dashboard when we list all the App

## Questions

- [x] What do we do with rules directly tied to sensitive data?
    - We keep them and the logic, considering it’s isolated to the code scan it won’t pause any problems.

cfabianski:
## Description

Closes #1468 

## Checklist


elsapet (approved):
Looks good

PR #1474, opened 2024-01-25
fix: handle comments for java

cfabianski:
## Description

Because the node Type was `line_comment` we weren't matching against it for expected rules (nor for disabled rules)
That fixes it.

Another approach would be to have the language report the comment node types but that felt overkill for now.

## Checklist


elsapet:
Thanks @cfabianski 🙏

PR #1516, opened 2024-02-27
feat(rules): improve rule dx

### LINKED ISSUE #1486: Warn when a rule cannot be igored because it cannot be found
when skipping a rule that dosent exist the cli exits early

```
bearer scan . --skip-rule=foo
Analyzing codebase
⠏  (10828748/-) [12s]
Loading rules
Error: invalid rule IDs in only/skip option: foo
```

This should be changed to a warning as with findings that are no longer found

NB: Only rule should still function in this way

### LINKED ISSUE #1500: Bearer silently fails when it can't download any rules
## Description & Reproduction

If you try to run bearer in an airgapped environment it will silently fail to download the rules.  

## Expected Behavior

I expect that in this case an error message is shown so the user can check his connection or provide his own rules.

## Actual Behavior

Bearer does not log an error message. Instead it continues and generates an empty report with 0 findings.
The exit code is 0 in this case.
This could lead to the assumption that the own software does not have any bugs that bearer can detect and that it is "safe"

## Possible Fix

Fail with an error message if bearer was not able to load any rules (neither local nor the default ones). 

## Your Environment

- Operating System and version: Linux - Docker Image in Gitlab Runner
- Output of 'bearer version': 
```
bearer version: 1.38.0
```

gotbadger:
## Description

- Improve experience when skiping rules and when rules cannot be downloaded.
- Set exit code to failure when no rules are found when it would be otherwise expected.

## Related

- Close #1486
- Close #1500

## Checklist

### DIFF
--- a/internal/commands/process/settings/rules.go
+++ b/internal/commands/process/settings/rules.go
@@ -325,18 +325,21 @@ func validateRuleOptionIDs(
 			invalidRuleIDs = append(invalidRuleIDs, id)
 		}
 	}
-
+	var invalidSkipRuleIDs []string
 	for id := range options.SkipRule {
 		_, existsInDefinition := definitions[id]
 		_, existsInBuiltInDefinition := builtInDefinitions[id]
 
 		if !existsInBuiltInDefinition && !existsInDefinition {
-			invalidRuleIDs = append(invalidRuleIDs, id)
+			invalidSkipRuleIDs = append(invalidSkipRuleIDs, id)
 		}
 	}
 
+	if len(invalidSkipRuleIDs) > 0 {
+		output.StdErrLog(fmt.Sprintf("Warning: rule IDs %s given to be skipped but were not found", strings.Join(invalidSkipRuleIDs, ",")))
+	}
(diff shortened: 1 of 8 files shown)


internal/commands/process/settings/rules.go
```diff
@@ -325,18 +325,21 @@ func validateRuleOptionIDs(
...
 		_, existsInBuiltInDefinition := builtInDefinitions[id]
 
 		if !existsInBuiltInDefinition && !existsInDefinition {
-			invalidRuleIDs = append(invalidRuleIDs, id)
+			invalidSkipRuleIDs = append(invalidSkipRuleIDs, id)
 		}
 	}
 
+	if len(invalidSkipRuleIDs) > 0 {
+		output.StdErrLog(fmt.Sprintf("Warning: rule IDs %s given to be skipped but were not found", strings.Join(invalidSkipRuleIDs, ",")))
```
gotbadger, line 339: This is currently not muted by --quiet but think thats ok considering

didroe (approved):

PR #1558, opened 2024-04-08
fix: handle duplicate var names in assignment

elsapet:
## Description

We were hitting issues with overridden var names e.g. 

```
	t := template.New()
	t, err := t.Parse(tmpl)
```

## Checklist

### DIFF
--- a/internal/languages/golang/detectors/.snapshots/TestGoObjects-object_no_class
+++ b/internal/languages/golang/detectors/.snapshots/TestGoObjects-object_no_class
@@ -109,8 +109,6 @@ children:
                       id: 22
                       range: 6:2 - 6:6
                       content: user
-                      alias_of:
-                        - 20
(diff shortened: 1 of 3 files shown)


internal/languages/golang/detectors/.snapshots/TestGoObjects-object_no_class
```diff
@@ -109,8 +109,6 @@ children:
                       id: 22
                       range: 6:2 - 6:6
                       content: user
-                      alias_of:
```
elsapet, line 112: Not sure about the snapshot changes - @didroe any concerns?

didroe: I think this is ok, let's see if all the rules tests pass once it's merged

didroe (approved):

PR #1589, opened 2024-05-13
fix(python): update analyzer

elsapet:
## Description

Add `parenthesized_expression` and a few more cases to the analyzer

We now catch the following sorts of things: 

```python

# args list with parenthesized expression
with fileinput.input(files=(filepath), encoding="utf-8") as f:

# args as arrays
subprocess.run([unsafe, "exit 0"], shell=True, capture_output=True)
```

## Checklist

### DIFF
--- a/internal/languages/python/analyzer/analyzer.go
+++ b/internal/languages/python/analyzer/analyzer.go
@@ -150,7 +150,7 @@ func (analyzer *analyzer) analyzeBoolean(node *sitter.Node, visitChildren func()
 	return visitChildren()
 }
 
-func (analyzer *analyzer) analyzeKeywordArgumentOperation(node *sitter.Node, visitChildren func() error) error {
+func (analyzer *analyzer) analyzeKeywordArgument(node *sitter.Node, visitChildren func() error) error {
 	value := node.ChildByFieldName("value")
 
 	analyzer.builder.Alias(node, value)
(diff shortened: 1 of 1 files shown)


internal/languages/python/analyzer/analyzer.go
```diff
@@ -35,10 +35,12 @@ func (analyzer *analyzer) Analyze(node *sitter.Node, visitChildren func() error)
...
 		return analyzer.analyzeCall(node, visitChildren)
-	case "argument_list":
-		return analyzer.analyzeGenericOperation(node, visitChildren)
-	case "expression_statement":
+	case "argument_list", "expression_statement", "list", "tuple":
 		return analyzer.analyzeGenericOperation(node, visitChildren)
+	case "parenthesized_expression":
+		return analyzer.analyzeGenericConstruct(node, visitChildren)
+	case "keyword_argument":
+		return analyzer.analyzeKeywordArgumentOperation(node, visitChildren)
```
didroe: WDYT about dropping "Operation" from the name?

```suggestion
		return analyzer.analyzeKeywordArgument(node, visitChildren)
```

internal/languages/python/analyzer/analyzer.go
```diff
@@ -148,6 +150,27 @@ func (analyzer *analyzer) analyzeBoolean(node *sitter.Node, visitChildren func()
 	return visitChildren()
 }
 
+func (analyzer *analyzer) analyzeKeywordArgumentOperation(node *sitter.Node, visitChildren func() error) error {
```
didroe:
```suggestion
func (analyzer *analyzer) analyzeKeywordArgument(node *sitter.Node, visitChildren func() error) error {
```

didroe (approved):

PR #1601, opened 2024-05-17
fix: handle identifiers in subscript

elsapet:
## Description

If there is an identifier in the subscript, we want to track this back to its origin

With this change, we can now pick up both of these cases: 

```python
globals()[input()] # no identifier

x = input()
globals()[x] # identifier
```

## Checklist

### DIFF
--- a/internal/languages/python/analyzer/analyzer.go
+++ b/internal/languages/python/analyzer/analyzer.go
@@ -120,10 +120,20 @@ func (analyzer *analyzer) analyzeAttribute(node *sitter.Node, visitChildren func
 }
 
 // foo["bar"]
+// foo[x]
+// globals()[y]
 func (analyzer *analyzer) analyzeSubscript(node *sitter.Node, visitChildren func() error) error {
 	objectNode := node.ChildByFieldName("value")
-	analyzer.builder.Dataflow(node, objectNode)
-	analyzer.lookupVariable(objectNode)
+	if objectNode.Type() == "identifier" {
+		analyzer.builder.Dataflow(node, objectNode)
+		analyzer.lookupVariable(objectNode)
+	}
+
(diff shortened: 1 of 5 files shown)


internal/languages/python/analyzer/analyzer.go
```diff
@@ -120,10 +120,20 @@ func (analyzer *analyzer) analyzeAttribute(node *sitter.Node, visitChildren func
...
-	analyzer.builder.Dataflow(node, objectNode)
-	analyzer.lookupVariable(objectNode)
+	if objectNode.Type() == "identifier" {
+		analyzer.builder.Dataflow(node, objectNode)
+		analyzer.lookupVariable(objectNode)
+	}
+
+	subscriptNode := node.ChildByFieldName("subscript")
+	if subscriptNode.Type() == "identifier" {
+		analyzer.builder.Dataflow(node, subscriptNode)
```
didroe: I don't think we want to create an association with the index/key

internal/languages/python/analyzer/analyzer.go
```diff
@@ -120,10 +120,20 @@ func (analyzer *analyzer) analyzeAttribute(node *sitter.Node, visitChildren func
 }
 
 // foo["bar"]
+// foo[x]
+// globals()[y]
 func (analyzer *analyzer) analyzeSubscript(node *sitter.Node, visitChildren func() error) error {
 	objectNode := node.ChildByFieldName("value")
-	analyzer.builder.Dataflow(node, objectNode)
-	analyzer.lookupVariable(objectNode)
+	if objectNode.Type() == "identifier" {
```
didroe: I think we want to do this unconditionally, `lookupVariable` will check the type anyway, and we'll want the dataflow association even if it's not a variable/identifier

didroe (approved):

PR #1603, opened 2024-05-17
fix: ensure parent folder symlinks are resolved

didroe:
## Description

Handle parent folder symlinks in `CanonicalPath`. 

We resolve symlinks but if the links are in an unspecified part of the path, then they aren't resolved currently. This is causing a mismatch with the target path and the paths returned from git.

## Checklist


cfabianski (approved):
💌

PR #1610, opened 2024-05-23
fix(python): handle string line continuation

didroe:
## Description

Workaround line continuation issue with Python strings.

Python string line continuation is incorrectly represented as an escape character by the tree sitter parser. This causes the detector to fail when trying to interpret it.

## Checklist


elsapet (approved):
🙏

PR #1622, opened 2024-06-05
fix(python): handle reflexive methods

elsapet:
## Description

Handle reflexive method function calls
Remove scoping from block

## Checklist


didroe (approved):

pkg/languages/python/detectors/.snapshots/TestPythonObjects-object_class
```diff
@@ -23,6 +23,9 @@ children:
         - type: block
           id: 5
           range: 2:5 - 8:33
+          dataflow_sources:
+            - 6
+            - 40
```
elsapet, line 28: @didroe heads up some snapshot changes around dataflow sources - I think they're fine?

didroe (approved):

PR #1623, opened 2024-06-05
fix(python): import statements

elsapet:
## Description

Following my changes here https://github.com/Bearer/bearer/pull/1622 we are no longer catching this kind of import: 

```
from hashlib import md5
md5(current_user.password) # not catching
```

## Checklist


didroe (approved):
Oops, sorry I think I told you we didn't need that

PR #1661, opened 2024-07-30
fix: remove use of rego keyword

didroe:
## Description

Renames `contains` to `array_contains` in rego code. 

`contains` is a Rego keyword and this causes an error in later versions. See https://github.com/Bearer/bearer/pull/1656

## Checklist


elsapet (approved):
🙌

PR #1803, opened 2025-07-29
feat: add option to not process gitignore files

gotbadger:
## Description
Address and edge case where files added before a git ignore is committed can exist in the project but not be scanned.

This option can also be used to make sure you scan every file present on disk, for example, when working with generated or built files that are not otherwise committed to source control.

### DIFF
--- a/docs/_data/bearer_scan.yaml
+++ b/docs/_data/bearer_scan.yaml
@@ -3,7 +3,7 @@ synopsis: Scan a directory or file
 usage: bearer scan [flags] <path>
 options:
   - name: api-key
-    usage: Use your Bearer API Key to send the report to Bearer.
+    usage: Legacy.
     environment_variables:
       - BEARER_API_KEY
   - name: config-file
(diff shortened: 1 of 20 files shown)


pkg/commands/process/orchestrator/worker/worker.go
```diff
@@ -37,12 +37,14 @@ type Worker struct {
...
 	sastScanner     *scanner.Scanner
 	skipTest        bool
+	skipGitIgnore   bool
 }
 
 func (worker *Worker) Setup(config config.Config) error {
 	worker.debug = config.Debug
 	worker.enabledScanners = config.Scan.Scanner
 	worker.skipTest = config.Scan.SkipTest
+	worker.skipGitIgnore = config.Scan.SkipTest
```
didroe:
```suggestion
	worker.skipGitIgnore = config.Scan.SkipGitIgnore
```

docs/_data/bearer_scan.yaml
```diff
@@ -160,6 +160,12 @@ options:
     usage: Specify which severities are included in the report.
     environment_variables:
       - BEARER_SEVERITY
+  - name: skip-git-ignore
+    default_value: "false"
+    usage: |
+      Do not automatically skip files that match patterns in .gitignore
```
elsapet: Suggestion: it's a bit clearer to me than a "do not" that I must set to `true` to enable. Take it or leave it, just an alternative option. 

```suggestion
      Scan files even if their paths match patterns in .gitignore
```

didroe (approved):

elsapet (approved):

PR #1876, opened 2025-12-16
feat(python): add dataflow for default parameter values

ronens88:
- analyzer.go: Create alias from parameter name to default value, enabling string resolution through function parameters
- filters.go: Return matched content from regex filter instead of empty string, allowing variable references to be captured

Enables tracing values through Python function default parameters.

## Description

This PR adds dataflow tracking for Python function parameters with default values.

**Problem:** Previously, when a function parameter with a default value was used inside the function body, Bearer couldn't trace it back to the default value string.

**Solution:** 
1. In `analyzer.go`, create an alias from the parameter name to its default value during parameter analysis
2. In `filters.go`, return the matched node content from the `regex` filter instead of an empty string

**Example - Before:**
def connect(host: str = "localhost"):
    return open_connection(host=host)  # host couldn't be resolved**Example - After:**
def connect(host: str = "localhost"):
    return open_connection(host=host)  # host now resolves to "localhost"This enables rules to detect string values passed through function parameters when they originate from default values.

## Checklist

### DIFF
--- a/pkg/scanner/detectors/customrule/filters/filters.go
+++ b/pkg/scanner/detectors/customrule/filters/filters.go
@@ -351,7 +351,8 @@ func (filter *Regex) Evaluate(
 	patternVariables variableshape.Values,
 ) (*Result, error) {
 	node := patternVariables.Node(filter.Variable)
-	result := filter.Regex.MatchString(node.Content())
+	content := node.Content()
+	result := filter.Regex.MatchString(content)
 
 	if log.Trace().Enabled() {
 		log.Trace().Msgf(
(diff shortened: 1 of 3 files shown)


pkg/scanner/detectors/customrule/filters/filters.go
```diff
@@ -351,19 +351,20 @@ func (filter *Regex) Evaluate(
...
 			"filters.Regex: %t for pattern %s at %s, content=%s",
 			result,
 			filter.Regex.String(),
 			node.Debug(),
-			node.Content(),
+			content,
 		)
 	}
 
-	return boolResult(patternVariables, result, ""), nil
```
cfabianski, line 366: Why do you need this?

cfabianski (approved):

PR #1984, opened 2026-06-23
chore(deps-dev): upgrade eleventy 2->3 and markdown-it 13->14 in /docs

elsapet:
Supersedes dependabot PR #1976 (which bumps only the two packages and would not build under eleventy v3 without the config/CI changes below).

## What

- `@11ty/eleventy` 2.0.1 → **3.1.6**
- `markdown-it` 13.0.2 → **14.2.0**
- `docs/.eleventy.js`: config callback made `async`; `EleventyHtmlBasePlugin` / `EleventyRenderPlugin` moved to a dynamic `import()` inside it
- CI Node 16 → **20** in `docs.yml` (build + Pages deploy) and `lint_docs.yml` (lint)

## Why the config change

Eleventy v3 publishes its named exports as ESM-only, so the top-level `require("@11ty/eleventy")` throws on Node < 22.12. Rather than convert the whole CommonJS config (and all the local `require`d modules) to ESM, only those two exports are dynamic-`import()`ed inside the now-`async` callback — the eleventy-recommended minimal fix.

## Why Node 20

Eleventy v3 requires Node ≥18; `gitly` wants ≥20. Node 20 (active LTS) satisfies both.

## Minimal churn

`syntaxhighlight` (4.2.0), `eleventy-fetch` (3.0.0), `eleventy-plugin-toc` (1.1.5) and `markdown-it-emoji` (2.0.2) are all compatible with v3 unchanged — no extra bumps, no markdown-it-emoji v3 import migration needed.

## Verification

- Clean `npm ci` + `npm run build` on Node 20-equivalent → exit 0, **880 files** on **v3.1.6** (identical count to v2), no warnings/errors
- `npm run lint` → exit 0
- Rendered `_site` checked: syntax highlighting (491 pages), mermaid (1), TOC (498) all render and coexist; **0 unrendered template/shortcode tags**

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version: "20"
+          node-version: "24"
       - name: Cache dependencies
         uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
         with:
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version: "16"
+          node-version: "20"
       - name: Cache dependencies
         uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/docs.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          node-version: "16"
+          node-version: "20"
```
gotbadger: lets go to 24 since 20 is deprecated

gotbadger (approved):

docs/_data/rules.js
```diff
@@ -82,13 +82,13 @@ async function fetchData(location) {
   const gitleaks = await fetchFile(gitleaksInternalRule, "/")
   rules.push(gitleaks)
   // ex: looping through rules [ruby, gitleaks, sql]
-  dirs.forEach(async (dir) => {
+  for (const dir of dirs) {
```
elsapet, line 85: Long-standing race bug that is now more exposed by eleventy-3 which processes the files faster (apparently)

PR #1985, opened 2026-06-25
feat: refactor ignore logic

gotbadger:
## Description
refactor ignore logic and fix case where test files would still be ignored after flag was set

## Checklist


elsapet (approved):
Neat solution 👍

